PR #2, opened 2015-04-17
Better insert update delete

sbosley:
This allows us to return values from the grammar versions of insert/update/delete that are more useful than just a boolean success or failure.


jdkoren:
LGTM

PR #3, opened 2015-04-17
Better locking mechanism

sbosley:
@sciolizer helpfully pointed out to me that I had re-invented a read/write lock with my locking mechanism for protecting attached databases. This replaces my cruder version of that locking with a ReentrantReadWriteLock, which is much simpler to understand and maintain.

Note: it's important to remember that the lock is not actually protecting database reads and writes (both of those can use the read lock)--it's only for protecting exclusive access to the database connection for things like attaches. For that reason I've left the original method naming of "exclusive" and "non-exclusive" locks and documented how each should be used.


jdkoren:
LGTM

PR #7, opened 2015-04-29
AbstractDatabase methods shouldn't suppress any SQLException

sbosley:
We used to suppress SQLExceptions in various places, but we weren't very consistent about it. I think that's the wrong approach--they're runtime exceptions, so they're not expected to be thrown unless something has gone badly wrong or the framework is being used incorrectly. Better to throw the exception than silently continue in a weird state.

Let's instead allow all SQLExceptions to be thrown, which should make debugging significantly easier. The "try" methods can still catch them and return booleans, but I've also added "execSqlOrThrow" as a counterpart to tryExecSql.


jdkoren:
LGTM

PR #8, opened 2015-04-30
Remove reflection

sbosley:
This PR removes the reflective lookups of the PROPERTIES and TABLE fields of model classes in favor of more direct access. This is achieved through a tweak in code generation that passes the properties array for a model to the table it constructs for that model. Conceptually this seems like it makes sense--a table has knowledge of the properties corresponding to it, and those properties have a pointer back to their parent table.

Api changes:
Property<?>[] parameters added to all SqlTable constrcutors
ContentProviderQueryBuilder constructor now takes a Property<?>[] and a SqlTable<?> parameter instead of a model class
The model class constructor for UriNotifier has been removed (the SqlTable constructor should be used instead)

Generally speaking we won't suddenly change apis like this. This is an exception we're making since most users won't notice and the library hasn't been in use for very long.

This is somewhat unscientific, but in my tests this change was actually a decent speedup--a 10% running time reduction in the unit test suite.


squidb/src/com/yahoo/squidb/data/UriNotifier.java
```diff
@@ -108,7 +96,8 @@ public UriNotifier(SqlTable<?>... tables) {
      * @param operation the type of database write that occurred
      * @param modelValues the model values that triggered this database update. This parameter may be null; the dao
      * will provide it when possible, but it is not always present. If you only need a row id, check the rowId
-     * parameter. This parameter will be null for delete operations, and will contain only the changed columns and their
+     * parameter. This parameter will be null for delete operations, and will contain only the changed columns and
+     * their
      * new values for updates.
```
jdkoren: bad Android Studio

squidb/src/com/yahoo/squidb/sql/Table.java
```diff
@@ -19,29 +18,32 @@
...
+            String alias) {
+        super(modelClass, properties, name);
         this.alias = alias;
         this.tableConstraint = tableConstraint;
     }
 
-    private Table(Class<? extends TableModel> modelClass, String name, String tableConstraint, String alias,
+    private Table(Class<? extends TableModel> modelClass, Property<?>[] properties, String name, String tableConstraint,
+            String alias,
             String databaseName) {
```
jdkoren: nit: line break (bad Android Studio)

jdkoren:
LGTM. Minor nits caused by Android Studio

PR #14, opened 2015-05-04
Better typesafety for ContentValues constructors

sbosley:
Our type safety when constructing models from ContentValues was lacking. This PR improves the situation:

There are now two constructors generated in each model class for constructing from ContentValues. One takes the values and a list of properties to read from the values, which is used to type-safely read each value from the ContentValues--if anything is of the wrong type or can't be coerced, a ClassCastException will be thrown. The other constructor just takes a single ContentValues and calls the two-arg constructor with the PROPERTIES array for that model.

This PR also increases type safety by making ValueCastingVisitor explicitly throw a ClassCastException when type coercion fails. We used to rely on the cast to (TYPE) in the get() method to throw this, but that won't work when get() is called with a property of unknown type (e.g. Property<?>).


jdkoren:
LGTM

PR #18, opened 2015-05-07
Fix defaultValues bug

sbosley:
Fix a bug where default values wouldn't get the ValueCastingVisitor, which could cause problems for boolean properties with integer defaults. This fixes a regression introduced in #14.

Long term we should really move to something better/more typesafe for our defaults, like separate annotations for them.


smb510:
LGTM! :goat:

jdkoren:
LGTM

PR #19, opened 2015-05-07
Fix gradle apt issues

sbosley:
I was making some tweaks to the annotation processor and all the gradle builds started failing whenever I made any change. Couldn't run the tests, couldn't clean/recompile the projects, nothing. After much random experimentation, I noticed that we still had some old precompiled jars of the processor, jackson plugin, etc. in the test project, so I removed those in case they were messing things up. Then I bumped the Android gradle plugin version to the latest in all the build.gradle files, and deleted my .gradle caches from the squidb directory AND my home directory. Once  I had completed that long ritual, things were working again as expected and I could make changes to the annotation processor without breaking things.


jdkoren:
LGMT

PR #21, opened 2015-05-08
Fix validation race condition

sbosley:
The 'parenthesizeWhere' approach we took when checking queries that needed validation had a race condition that could cause SQL syntax errors. Specifically, this could arise if there was a subquery shared by many other queries--e.g., exactly what happens when using ContentProviderQueryBuilder with a subquery ViewModel. If queries were being executed on multiple threads, validation on thread A could parenthesize the where clause of the shared subquery while thread B did the same. Thread A could then unset the parenthesize where flag before thread B was done compiling the query, resulting in unbalanced parentheses for thread B. 

Note: while this problem could theoretically apply to any Query that has requestValidation called on it, the only place where SquiDB uses it is in ContentProviderQueryBuilder, so it's very unlikely that users not using CPQB would encounter this problem.

This PR fixes the issue by removing the statefulness of the parenthesizeWhere flag and just passing a boolean argument through to the relevant methods when compiling Query objects into SQL. This should allow multiple threads to validate queries that have shared state without interfering with each other.

I'm pretty sure that this should fix the problem and is the correct approach, but I don't have any great ideas for new unit tests to confirm at the moment. Let's hold off on merging this PR until I can add at least one or two new ones to ensure things are definitely working as expected.

UPDATE: added a unit test based on a suggestion by @jdkoren. The new test fails on master by forcing the race condition, but passes on this PR.

### DIFF
--- a/squidb-tests/src/com/yahoo/squidb/sql/QueryTest.java
+++ b/squidb-tests/src/com/yahoo/squidb/sql/QueryTest.java
@@ -23,6 +23,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.Semaphore;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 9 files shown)


jdkoren:
LGTM

Issue #23, opened 2015-05-28
Enhance rowid handling for virtual tables


sbosley:
Right now, rowid handling for virtual tables is inconsistent with how rowids are handled for regular tables. There are two core issues:
 - [x] Calling setId() explicitly on a virtual table model will break its ability to be persisted, since setId() uses the column name "_id" but virtual tables require "rowid". (Things will still work fine if the id was read from a cursor result).
 - [ ] Queries of the form "select *" don't return the virtual table rowids, meaning the row id won't be read into the models unless a specific list of properties is specified. This applies to query methods and all fetch methods.

We should figure out how to best make virtual table models behave like regular table models in a way that is transparent to the user.

PR #24, opened 2015-05-28
Tables and virtual tables

sbosley:
This PR makes tables and virtual tables more interchangeable by refactoring VirtualTable to make it a subclass of table. This required a few tweaks to the code generator to make the constructor arguments play nice. Both regular tables and virtual tables are now provided to the database by the getTables() method, so getVirtualTables() has been removed.

There are still some ways in which tables and virtual tables are not entirely interchangeable, most notably in their handling of the rowid column. Specifically, select * queries on virtual tables don't return the rowid, and calling setId on a virtual table model breaks its ability to be persisted correctly. Issue #23 has been opened for these to be investigated separately.


squidb/src/com/yahoo/squidb/data/AbstractDatabase.java
```diff
@@ -49,7 +50,8 @@
  * recommended for type safety reasons. Instead, use an instance of {@link DatabaseDao} to issue the request and return
  * a {@link SquidCursor}.
  * <p>
- * By convention, methods beginning with "try" (e.g. {@link #tryCreateTable(Table) tryCreateTable}) return true if the
+ * By convention, methods beginning with "try" (e.g. {@link #tryCreateTable(ConcreteTable) tryCreateTable}) return true
+ * if the
```
jdkoren: grrr Studio...

jdkoren:
LGTM

PR #25, opened 2015-05-28
More ContentValues typesafety

sbosley:
A few weeks ago we merged #14 to improve the type safety when reading ContentValues into a model object. I discovered that we had another method that was similar--AbstractModel.setValues(). It put values from a ContentValues into the setValues non-typesafely. This PR introduces setPropertiesFromContentValues as an analogue to readPropertiesFromContentValues and removes the bad setValues method.


jdkoren:
LGTM

PR #26, opened 2015-05-29
Add the ability to build artifacts for deploying to jcenter

sbosley:
Add code to our build.gradle files for building artifacts that we can deploy to a maven repo. Once this is merged, I'll deploy the 1.0.0 artifacts and update the issues/wiki with how you can build from artifacts instead of cloning and building from source.

I also fixed a few things in the javadoc that were throwing warnings when built by the maven javadoc builder.


jdkoren:
LGTM

jdkoren:
oops, wrong button

PR #33, opened 2015-06-03
SquiDB sample project

### LINKED ISSUE #11: Hello SquiDB sample project
Can you prepare a simple "hello world" sampel project that will show base SquiDB features?

e.g. https://github.com/yahoo/squidb/wiki/Setting-up-a-database-for-your-app should contain an example of the database and the usage - I've found it on the readme.md but still this would be useful as it is hard to collect all necessary elements to make working app.

sbosley:
A basic Hello SquiDB sample project that shows off some basic project setup, model specs, query usage, custom sqlite functions, uri notifiers for listening to data changes, etc.

The sample app itself implements a (very basic) todo list. You can insert tasks (tagged with an arbitrary number of tags), complete tasks, and delete tasks. The sample app is very simple; we will probably occasionally enhance it to show off more advanced features in the future.

Closes #11.


LukaszGrela:
Thanks:)
I was preparing something similar:) https://github.com/LukaszGrela/TodoOrNotTodo

LukaszGrela:
Can you explain the use of this dependencies?

    compile 'javax.inject:javax.inject:1'
    compile 'com.squareup.dagger:dagger:1.2.2'
    apt 'com.squareup.dagger:dagger-compiler:1.2.2'

Could the sample be stripped off any 3rd parties and only plain SquiDB framework used? Just to avoid confusion:)

sbosley:
That's [Dagger](http://square.github.io/dagger/), Square's dependency injector. It's used in the sample project as an example of how to make the DatabaseDao and AbstractDatabase instances singletons. We could remove it and use a static instance to provide a singleton, but static instances aren't always great practice, particularly when considering unit testing or in larger projects. And while I suppose neither of those considerations apply to the sample project :smile: I figured more people would be using a dependency injector of some kind than not.

I can see why it would be desirable for the sample project to be clean of dependencies though...I'll think about it. At the very least, we could add in some more documentation explaining the dependency and why it's used.

LukaszGrela:
yes please, then definitely I will benefit from new knowledge:D

PR #37, opened 2015-06-17
Primary key enhancements

### LINKED ISSUE #27: Change default id name
I have some pre-populated sqlite databases that I'd like to access via squidb. The id columns on all of my tables are named "id" (not "_id"). I'd like to be able to change the default id name in squidb to "id" for this reason.

sbosley:
This PR introduces a new annotation, PrimaryKey, which will allow users to customize their column definitions for primary keys. At this time, only INTEGER PRIMARY KEYs are supported, but the following things are now possible:
 * Name the primary key column anything you want. Helps with backwards compatibility and closes #27 
 * Disable the AUTOINCREMENT behavior which was previously required
 * Fixes setId behavior for virtual table models (half of #23)

I think this PR also lays some groundwork for multi-column primary key support in the future, but I'm not quite ready to tackle that.


jdkoren:
LGTM

Issue #42, opened 2015-07-07
Failed upgrades/downgrades can suppress exceptions


jdkoren:
Currently if a database upgrade or downgrade fails, a call to `onMigrationFailed()` is made to give an opportunity to take corrective action. Because this is not an abstract method and the default implementation is a noop, this effectively suppresses exceptions by default.

sbosley:
Closed by #43

PR #43, opened 2015-07-07
Don't suppress exceptions from upgrade/downgrade

jdkoren:
Fixes Issue #42 (github.com/yahoo/squidb/issues/42)


sbosley:
This is great. It also allows users to pass more detailed information about migration failures to themselves if they want to by subclassing MigrationFailedException and throwing it directly from onUpgrade. Do you think we could document that option as a valid use before we merge?

sbosley:
LGTM

PR #48, opened 2015-07-22
Fix issues regarding attaching databases with WAL enabled

jdkoren:
On API 16 and later, attempting to attach a database throws an exception if
* The attacher has WAL enabled; and
* The attacher has any open transactions on other threads

This situation is addressed by acquiring the exclusive lock to run the ATTACH command. Additionally, `attachDatabase()` will now throw an exception if the attacher has an open transaction on the current thread (whereas previously it just returned null).

### DIFF
--- a/squidb/src/com/yahoo/squidb/sql/VirtualTable.java
+++ b/squidb/src/com/yahoo/squidb/sql/VirtualTable.java
@@ -5,6 +5,8 @@
  */
 package com.yahoo.squidb.sql;
 
+import android.os.Build;
+
 import com.yahoo.squidb.data.AbstractDatabase;
 import com.yahoo.squidb.data.TableModel;
 import com.yahoo.squidb.sql.Property.PropertyVisitor;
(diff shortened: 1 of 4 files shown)


squidb/src/com/yahoo/squidb/sql/VirtualTable.java
```diff
@@ -72,8 +74,11 @@ public String toString() {
      */
     @Override
     public void appendCreateTableSql(StringBuilder sql, PropertyVisitor<Void, StringBuilder> propertyVisitor) {
-        sql.append("CREATE VIRTUAL TABLE IF NOT EXISTS ").append(getExpression()).append(" USING ").append(moduleName)
-                .append('(');
+        sql.append("CREATE VIRTUAL TABLE ");
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) {
```
sbosley: Good catch, I had noticed this yesterday as well. We'll need to revisit this if we ever decide to allow bundling other versions of SQLite, but for now this is good.

sbosley:
LGTM. Any other changes you plan to make on this branch or shall we merge?

jdkoren:
Nope, just wanted to make sure you reviewed it.

sbosley:
Cool, merging.

PR #54, opened 2015-07-27
ICS fixes

jdkoren:
Fixes some of the issues that occur on Ice Cream Sandwhich

### DIFF
--- a/squidb/src/com/yahoo/squidb/data/SqlValidatorFactory.java
+++ b/squidb/src/com/yahoo/squidb/data/SqlValidatorFactory.java
@@ -21,7 +21,7 @@ import com.yahoo.squidb.data.SqlValidatorFactory.SqlValidator;
         if (version >= VERSION_CODES.JELLY_BEAN) {
             INSTANCE = new DefaultSqlValidator();
         } else if (version >= VERSION_CODES.ICE_CREAM_SANDWICH) {
-            INSTANCE = new ReflectiveSqlValidator();
+            INSTANCE = new IcsSqlValidator();
         } else {
             // included for forks that reduce minSdk below 14
             INSTANCE = new DefaultSqlValidator();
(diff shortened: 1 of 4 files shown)


squidb/src/com/yahoo/squidb/data/SqlValidatorFactory.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    private static final SqlValidator INSTANCE;
+
+    static {
+        int version = VERSION.SDK_INT;
+        if (version >= VERSION_CODES.JELLY_BEAN) {
+            INSTANCE = new DefaultSqlValidator();
+        } else if (version >= VERSION_CODES.ICE_CREAM_SANDWICH) {
+            INSTANCE = new ReflectiveSqlValidator();
```
sbosley: This probably shouldn't be named ReflectiveSqlValidator now that it's no longer using a reflective approach.

squidb/src/com/yahoo/squidb/data/SqlValidatorFactory.java
```diff
@@ -0,0 +1,57 @@
...
+    private SqlValidatorFactory() {
+        //no instance
+    }
+
+    static SqlValidator getValidator() {
+        return INSTANCE;
+    }
+}
+
+/*package*/ class DefaultSqlValidator implements SqlValidator {
```
sbosley: Do these need to be package or can they be private static?

squidb/src/com/yahoo/squidb/data/SqlValidatorFactory.java
```diff
@@ -0,0 +1,56 @@
...
+        public void compileStatement(SQLiteDatabase db, String sql) {
+            db.compileStatement(sql);
+        }
+    }
+
+    private static class IcsSqlValidator implements SqlValidator {
+
+        @Override
+        public void compileStatement(SQLiteDatabase db, String sql) {
+            Cursor c = db.rawQuery(sql, null);
```
sbosley: Does this work even when there are no bound arguments? I'm assuming yes, because the unit tests are passing, but just making sure.

sbosley:
LGTM

PR #56, opened 2015-07-28
Add support for declaring that a model implements interfaces

### LINKED ISSUE #22: Model Interfaces
Is there any way to generate a model that implements an interface?

sbosley:
Closes #22

### DIFF
--- a/squidb-annotations/src/com/yahoo/squidb/annotations/Implements.java
+++ b/squidb-annotations/src/com/yahoo/squidb/annotations/Implements.java
@@ -12,12 +12,12 @@ import java.lang.annotation.Target;
  * This annotation can be used to declare that a generated model class should implement the listed interfaces. This
  * only affects the <code>implements X, Y, Z</code> declaration of the generated class--you will still have to ensure
  * that the generated class implements any required methods yourself using {@link ModelMethod}
- * <br><br>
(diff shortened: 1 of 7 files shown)


squidb-annotations/src/com/yahoo/squidb/annotations/Implements.java
```diff
@@ -0,0 +1,68 @@
...
+         * itself. This argument will be ignored if there are any type argument classes specified using
+         * {@link #interfaceTypeArgs()}
+         */
+        String[] interfaceTypeArgNames() default {};
+    }
+
+    /**
+     * For specifying interface classes that the generated model should declare it implements
+     */
+    Class<?>[] interfaceClasses() default {};
```
jdkoren: I think it might be more clear if we take this out and accept just a list of `InterfaceSpec`s. `interfaceTypeArgs` and `interfaceTypeArgNames` could just be empty for interfaces that don't have a type argument.

sbosley: It's way more verbose though. I would imagine that most interfaces don't have generic args, so it's a lot nicer to be able to just specify class names.

jdkoren:
LGTM

PR #59, opened 2015-07-29
Add a travis build matrix to run tests on multiple emulator versions

sbosley:
I'm not totally sure that I've done this correctly, but opening the PR should kick off a build to tell us!


jdkoren:
LGMT

PR #60, opened 2015-07-29
Hopefully fix Travis timeout issues

sbosley:
Travis builds seem to occasionally fail with a timeout installing the APK. This variable increases the timeout threshold


jdkoren:
LGTM

PR #64, opened 2015-07-30
Random fixes

sbosley:
None of these are super critical, they're mostly just micro optimizations.
* Handle AtomicBooleans at argument binding time rather than treating them as reference arguments like we already do for things like AtomicLong/Integer (which are subclasses of Number).
* SquidCursorAdapter.hasStableIds should return true by default.
* Use the zeroblob() function for escaping null characters in inlined strings rather than blob literals. Supposedly SQLite handles zeroblobs very efficiently.


jdkoren:
LGMT

PR #67, opened 2015-08-04
Unified SquidDatabase [2.0]

sbosley:
This is a large api change intended for SquiDB 2.0. The basic idea is to merge the responsibilities of AbstractDatabase and DatabaseDao into a single, unified class called SquidDatabase. This has several benefits:

* The fact that DatabaseDao wrapped an AbstractDatabase subclass made it hard to know what kind of database you were talking to, especially in apps with multiple databases. Now, all that's required is to subclass SquidDatabase instead of AbstractDatabase--that part of the api remains the same. But now, you can now talk directly to your SquidDatabase subclass for all database I/O, which means you always know what kind of database you're talking to.
* You can define custom database access methods directly in your database subclass using dao methods without also needing to create a subclass of DatabaseDao
* Dao methods that work with model objects are now easily available from hooks like onUpgrade() and onTablesCreated().

Further changes intended for SquiDB 2.0 will build on this foundation.


squidb-sample/src/main/java/com/yahoo/squidb/sample/modules/HelloSquiDBModule.java
```diff
@@ -41,20 +40,19 @@ Context provideContext() {
...
+    TasksDatabase provideTasksDatabase(@ForApplicaton Context context) {
+        TasksDatabase database = new TasksDatabase(context);
 
         // Setting up a UriNotifier will sent ContentObserver notifications to Uris on table writes
-        dao.registerUriNotifier(new UriNotifier(Task.TABLE, Tag.TABLE) {
+        database.registerUriNotifier(new UriNotifier(Task.TABLE, Tag.TABLE) {
             @Override
             public void addUrisToNotify(Set<Uri> uris, SqlTable<?> table, String databaseName,
-                    DBOperation operation, AbstractModel modelValues, long rowId) {
+                                        DBOperation operation, AbstractModel modelValues, long rowId) {
```
jdkoren: weird whitespace addition

squidb/src/com/yahoo/squidb/data/SquidDatabase.java
```diff
@@ -58,24 +68,39 @@
...
+ * found, or null if no records are found for that particular form of fetch.
  */
-public abstract class AbstractDatabase {
+public abstract class SquidDatabase {
 
     /**
      * @return the database name
      */
-    protected abstract String getName();
+    public abstract String getName();
```
jdkoren: Just curious, but any reason this became public?

sbosley: getDatabasePath() is public, so I figured why not make getName() public. I don't think it's a bad idea that a SquidDatabase can publicly expose it's name. I'm not wedded to it though if you think we should revert it.

squidb/src/com/yahoo/squidb/utility/SquidUtilities.java
```diff
@@ -97,8 +97,6 @@ private static void copyFile(File in, File out) throws Exception {
         FileOutputStream fos = new FileOutputStream(out);
         try {
             copyStream(fis, fos);
-        } catch (Exception e) {
-            throw e;
```
jdkoren: Haha, wow

sbosley: I bet that's been there since Astrid :)

squidb/src/com/yahoo/squidb/data/SquidDatabase.java
jdkoren: I don't see any harm in it being public, was more wondering if something in the codebase required it.

jdkoren:
LGTM

PR #68, opened 2015-08-05
Misc enhancements

sbosley:
* Adds the ability to wrap a Query in a Function, so you can put a subquery in a select clause or construct criterions from it
* Adds a ProjectionMap constructor that copies an existing ProjectionMap


jdkoren:
LGTM

PR #69, opened 2015-08-07
Query validation fixes [2.0]

sbosley:
This PR makes query validation more robust. 

Now that queries can be wrapped in functions and so appear anywhere a Field is valid (already on the dev_2.0 branch), it became too difficult to keep track of the needsValidation flag at the Query level. Instead, a new SqlBuilder class was introduced to wrap all the objects necessary for compiling a SQL grammar object--the StringBuilder, the bound arguments List, and a new flag that indicates if any subquery of the root query was flagged as needing validation. When executing a query, we first compile the query and check to see if the flag is set in the CompiledStatement object before executing it. If the flag is set, it's at that point we do query validation.


jdkoren:
LGTM. I like a lot of the cleanup that accompanies this change

PR #70, opened 2015-08-10
Universal data changed notifier [2.0]

sbosley:
This PR abstracts out the concepts from our UriNotifier to introduce a fully generic/customizable data changed notification mechanism. The abstract class DataChangedNotifier can be overridden to accumulate an arbitrary set of objects to notify based on data changes during transactions and send notifications to those objects at the end of a successful transaction. UriNotifier then becomes a specific instance of DataChangedNotifier--the objects accumulated are the Uris themselves, and sendNotification is implemented by calling ContentResolver.notifyChange.

SimpleDataChangedNotifier is also introduced as a very simple implementation of a DataChangedNotifier. It simply allows running arbitrary code at the end of a successful transaction if the tables it listens to have changed by overriding a single method.

API changes:
* SquidDatabase.registerUriNotifier and related methods have been renamed to things like registerDataChangedNotifier, etc.
* UriNotifier.addUrisToNotify has been renamed to accumulateNotificationObjects, since it is now a subclass of the generic DataChangedNotifier class.

### DIFF
--- a/squidb-sample/src/main/java/com/yahoo/squidb/sample/modules/HelloSquiDBModule.java
+++ b/squidb-sample/src/main/java/com/yahoo/squidb/sample/modules/HelloSquiDBModule.java
@@ -9,6 +9,7 @@ import android.content.Context;
 import android.net.Uri;
 
 import com.yahoo.squidb.data.AbstractModel;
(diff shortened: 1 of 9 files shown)


jdkoren:
LGTM

Issue #36, opened 2015-06-13
Investigate using the Android SQLite bindings project to support custom SQLite builds


alipov:
Are there any plans to support [sqlcipher][1] as an underlying database?

[1]: https://github.com/sqlcipher/sqlcipher

sbosley:
We've considered the possibility of leveraging the [Android SQLite bindings project](https://www.sqlite.org/android/doc/trunk/www/index.wiki) to allow using SquiDB with a version of SQLite other than the stock version Android ships with. I imagine the ability to use the sqlcipher extension would fall under that banner, although I don't know off the top of my head if there would be any pitfalls to using a SQLite extension rather than just a standard SQLite build. I think it's something we'd like to do at some point, but I can't commit to a definite timeline for adding it at the moment.

I'll update the issue title to reflect that we'd like to investigate the general ability to use a custom SQLite version, and we'll consider sqlcipher as one of the possible versions when we get to the point of actually working on this. Thanks for opening the issue!

jdkoren:
@alipov On our dev_2.0 branch we just merged a PR that introduces a `squidb-sqlite-bindings` module. This module demonstrates how to use a custom SQLite implementation packaged with the APK; the module itself uses the latest official SQLite release.

When SquiDB 2.0 is released, we will also be publishing this module as an artifact for gradle. Developers who want to use some other compiled version of SQLite can pull down the module and build from source. We'll include instructions in the wiki pages at that time.

Regarding SQLCipher, we ran into some problems trying to make a similar bindings module for it. The short explanation is the SQLCipher library classes are based on outdated versions of the `android.database` classes (they appear to be from API 10), and SquiDB relies on more recent versions for correct behavior (our minimum is API 14). The differences make it unfeasible and perhaps impossible to support it at this time.

Sorry we don't have better news than this. We really would like to support this however, so we plan to reach out to the SQLCipher developers about updating their classes to match the more recent revisions of Android's. If SQLCipher does receive some updates, we'll make another attempt to support it then.

alipov:
> On our dev_2.0 branch we just merged a PR that introduces a squidb-sqlite-bindings module.

Thanks for the update, I'll take a look.


> We really would like to support this however, so we plan to reach out to the SQLCipher developers about updating their classes to match the more recent revisions of Android's.

[...]

PR #74, opened 2015-08-12
Support custom sqlite builds [2.0]

sbosley:
This PR lays the foundation for using custom native SQLite builds in place of the version of SQLite bundled with Android and introduces the squidb-sqlite-bindings project.

From the perspective of the public API, very little has changed. The main difference is that any of the hooks in SquidDatabase that returned/used android.database.sqlite.SQLiteDatabase as a parameter now use a new SquiDB interface, SQLiteDatabaseWrapper. If you use/override these hooks (e.g. onUpgrade()) you'll have to update the method signature to use the new interface. The wrapper itself exactly duplicates the public API of Android's SQLiteDatabase, so it should be usable interchangeably. The wrapper allows us to swap in instances of things like org.sqlite.database.sqlite.SQLiteDatabase from the [Android SQLite Bindings Project](https://www.sqlite.org/android/doc/trunk/www/index.wiki) wrapped in an adapter class.

By default, SquidDatabase will still use the version of SQLite that ships with android and the standard SQLite classes in the android.database.sqlite namespace. However, subclasses can override a new method getDatabaseOpenHelper to return a DatabaseOpenHelper that can control binding to a different SQLite binary.

A squidb-sqlite-bindings project has been introduced that provides a standard stock SQLite build and the three small adapter classes to integrate it with SquiDB. This project targets API 16+. It's easiest to see how simple the API is in an example:

```java
public class MyDatabase extends SquidDatabase {
    ...

    @Override
    protected DatabaseOpenHelper getDatabaseOpenHelper(Context context, String databaseName,
            DatabaseOpenHelperDelegate delegate, int version) {
        return new SQLiteBindingsDatabaseOpenHelper(context, databaseName, delegate, version);
    }
}
```

That's it! By depending on the squidb-sqlite-bindings project and overriding this one method, the SquidDatabase will connect to the latest version of SQLite instead of the stock build shipped with Android.

In future work, we will introduce another small bindings project to facilitate using SQLCipher as an underlying database (#36).

***
This supersedes #73, which I accidentally opened against master instead of the dev_2.0 branch.

Other changes in this PR:
* We were starting to have a lot of add-ons that weren't part of SquiDB proper, so I moved them into a new squidb-addons folder so the project would be a little more organized.
* Updates our travis builds to run the unit tests twice on each api version tested--once using the stock Android SQLite build and once using a custom SQLite build from the squidb-sqlite-bindings project.


jdkoren:
LGTM

Issue #16, opened 2015-05-07
DatabaseDao - method to clear table


LukaszGrela:
I wasn't able to find a quick "clear table" method, I've found this to work in that fashion:

        Criterion everywhere = Todo.ID.isNotNull();
        app.todosDao().deleteWhere(Todo.class, everywhere);

is it useful wrapped as a public method of `DatabaseDao` 

    /**
     * Delete all rows in the given {@Link TableModel} table
     * 
     * @param modelClass the model class corresponding to the table to delete from
     * @return the number of rows deleted on success, -1 on failure
     */
    public int deleteAll(Class<? extends TableModel> modelClass){
        return deleteWhere(modelClass, TableModel.ID_PROPERTY.isNotNull());
    }

jdkoren:
You can use `databaseDao.deleteWhere(Model.class, Criterion.all);` to delete all the rows.

PR #75, opened 2015-08-12
Remove Criterion.all and Criterion.none [2.0]

sbosley:
Using literals for these criterions actually had some unexpected performance implications, and they weren't super useful anyways. This PR removes them and updates a few things to replace them:

* Passing a null criterion to methods like deleteWhere or update is now equivalent to "delete all" and "update all". Accordingly, new convenience methods deleteAll and updateAll have been added to SquidDatabase (refs #16, which was closed but will now be better off than it was when 2.0 launches :) )
* Passing null for a Criterion value is handled more gracefully in several places. It is ignored where possible (e.g. query.where(null) is a no-op) and throws an exception where it is very bad (e.g. Criterion.negate(null) is nonsensical, so throws an IllegalArgumentException).

Users who still want the old behavior of Criterion.all (i.e. "WHERE 1") can use Criterion.literal(1) instead.

### DIFF
--- a/squidb/src/com/yahoo/squidb/sql/Function.java
+++ b/squidb/src/com/yahoo/squidb/sql/Function.java
@@ -129,10 +129,10 @@ public abstract class Function<TYPE> extends Field<TYPE> {
     }
 
     /**
-     * Create a Function that counts all rows
+     * Create a Function that counts all rows (i.e. count(*))
      */
     public static Function<Integer> count() {
-        return new ArgumentFunction<Integer>("COUNT", 1);
+        return new RawFunction<Integer>("COUNT(*)");
     }
 
     /**
--- a/squidb/src/com/yahoo/squidb/sql/Function.java
+++ b/squidb/src/com/yahoo/squidb/sql/Function.java
@@ -246,6 +246,9 @@ public abstract class Function<TYPE> extends Field<TYPE> {
      * Begins a CASE statement, populating it with the first WHEN ... THEN branch
      */
     public static CaseBuilder caseWhen(Criterion when, Object then) {
+        if (when == null) {
+            throw new IllegalArgumentException("Can't construct a CASE WHEN statement with a null criterion");
+        }
         return new CaseBuilder(null).when(when, then);
     }
 
--- a/squidb-tests/src/com/yahoo/squidb/data/ModelTest.java
+++ b/squidb-tests/src/com/yahoo/squidb/data/ModelTest.java
@@ -7,7 +7,6 @@ package com.yahoo.squidb.data;
 
 import android.content.ContentValues;
 
-import com.yahoo.squidb.sql.Criterion;
 import com.yahoo.squidb.sql.Property;
 import com.yahoo.squidb.test.DatabaseTestCase;
 import com.yahoo.squidb.test.TestModel;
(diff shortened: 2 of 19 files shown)


squidb/src/com/yahoo/squidb/sql/Function.java
```diff
@@ -129,10 +129,10 @@ protected String separator() {
     }
 
     /**
-     * Create a Function that counts all rows
+     * Create a Function that counts all rows (i.e. count(*))
      */
     public static Function<Integer> count() {
-        return new ArgumentFunction<Integer>("COUNT", 1);
+        return new RawFunction<Integer>("COUNT(*)");
```
jdkoren: Not "count(1)"? (or does it not matter?)

squidb/src/com/yahoo/squidb/sql/Function.java
```diff
@@ -246,6 +246,9 @@ protected String separator() {
      * Begins a CASE statement, populating it with the first WHEN ... THEN branch
      */
     public static CaseBuilder caseWhen(Criterion when, Object then) {
+        if (when == null) {
+            throw new IllegalArgumentException("Can't construct a CASE WHEN statement with a null criterion");
```
jdkoren: Maybe it's better to move this to CaseBuilder.when() itself

squidb/src/com/yahoo/squidb/sql/Function.java
sbosley: From the SQLite docs: "The count(X) function returns a count of the number of times that X is not NULL in a group. The count(*) function (with no arguments) returns the total number of rows in the group."

So, I think count(1) and count(*) effectively end up equivalent, but a raw function that is just the literal count(*) seemed cleaner to me than binding a 1.

jdkoren: Interesting. For some reason i thought the count(1) was advantageous over count(*), but can't remember why...

squidb/src/com/yahoo/squidb/sql/Function.java
sbosley: Agreed, updated.

squidb/src/com/yahoo/squidb/sql/Function.java
sbosley: They should be equivalent. Couldn't find anything specifically mentioning SQLite, but http://stackoverflow.com/questions/1221559/count-vs-count1. IMHO count(*) is easier to understand as "cardinality", where as sometimes when I see count(1) I say "wait, what are the rules again?"

squidb/src/com/yahoo/squidb/sql/Function.java
sbosley: We did this and then realized we were wrong. "when null" is sometimes valid, e.g. "case x when null then 0 ..." So we only disallow null for criterions in the "case when <criterion>" case. If you really want "case when null", you can use Criterion.literal(null).

squidb/src/com/yahoo/squidb/sql/Function.java
jdkoren: Cool, thanks for clarity

jdkoren:
LGTM

PR #76, opened 2015-08-14
Assorted virtual model fixes

### LINKED ISSUE #23: Enhance rowid handling for virtual tables

sbosley:
This includes two changes:
1) A small tweak to the processor to disallow using the @PrimaryKey annotation in a virtual table model spec (since the primary key in a virtual table is always the rowid).
2) A small enhancement to Query that explicitly includes the table rowid when doing a select * from a virtual table. select * wouldn't include the rowid by default but we don't want to make models fetched from the database appear unpersisted.

Closes #23

### DIFF
--- a/squidb-processor/src/com/yahoo/squidb/processor/writers/ModelFileWriter.java
+++ b/squidb-processor/src/com/yahoo/squidb/processor/writers/ModelFileWriter.java
@@ -122,19 +122,20 @@ public abstract class ModelFileWriter<T extends Annotation> {
         }
         if (!modifiers.contains(Modifier.STATIC)) {
             utils.getMessager().printMessage(Kind.WARNING,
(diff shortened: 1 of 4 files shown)


jdkoren:
LGTM

PR #80, opened 2015-08-21
Misc 2.0 fixes

sbosley:
This fixes an issue with DBObject hashCode and equals--they relied on getExpression(), but Function and function properties don't support getExpression(). This fixes that bug as well as a few miscellaneous warnings.

### DIFF
--- a/squidb-tests/src/com/yahoo/squidb/data/PropertyTest.java
+++ b/squidb-tests/src/com/yahoo/squidb/data/PropertyTest.java
@@ -60,7 +60,7 @@ public class PropertyTest extends SquidTestCase {
         assertEquals(test1.hashCode(), test2.hashCode());
 
         StringProperty test3 = new StringProperty(TestModel.TABLE, "testCol");
-        StringProperty test4 = new StringProperty(TestModel.TABLE, "testCol", "DEFAULT 'A");
+        StringProperty test4 = new StringProperty(TestModel.TABLE, "testCol", "DEFAULT 'A'");
 
         assertEquals(test3, test4);
         assertEquals(test3.hashCode(), test4.hashCode());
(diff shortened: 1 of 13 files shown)


squidb/src/com/yahoo/squidb/sql/DBObject.java
```diff
@@ -75,27 +75,35 @@ public boolean equals(Object o) {
         if (alias != null ? !alias.equals(dbObject.alias) : dbObject.alias != null) {
             return false;
         }
-        String myExpression = getExpression();
-        String otherExpression = dbObject.getExpression();
-        if (myExpression != null ? !myExpression.equals(otherExpression) : otherExpression != null) {
+        if (expressionForComparison() != null ? !expressionForComparison().equals(dbObject.expressionForComparison())
```
jdkoren: I generally like to do the method call once, but not opposed to doing it this way.

squidb-tests/src/com/yahoo/squidb/data/PropertyTest.java
```diff
@@ -48,4 +52,30 @@ public void testPropertyAliasing() {
...
 
+    public void testEqualsAndHashCode() {
+        LongProperty test1 = new LongProperty(TestModel.TABLE, "testCol");
+        LongProperty test2 = new LongProperty(TestModel.TABLE, "testCol");
+
+        assertEquals(test1, test2);
+        assertEquals(test1.hashCode(), test2.hashCode());
+
+        StringProperty test3 = new StringProperty(TestModel.TABLE, "testCol");
+        StringProperty test4 = new StringProperty(TestModel.TABLE, "testCol", "DEFAULT 'A");
```
jdkoren: missing single quote?

squidb/src/com/yahoo/squidb/sql/DBObject.java
sbosley: Oh yeah, this was generated by Studio. I can update it

squidb-tests/src/com/yahoo/squidb/data/PropertyTest.java
sbosley: Yep. Doesn't matter for the purposes of the test but I will update.

PR #81, opened 2015-08-24
Gradle version flexibility

sbosley:
Apply the correct version of the android-maven gradle plugin for the current version of gradle. This allows people building from source to use whatever gradle version they want between 2.2 and 2.5 to build from source, since this plugin version was the only thing standing in the way of that.


jdkoren:
LGTM

PR #85, opened 2015-09-09
SquidRecyclerAdapter

jdkoren:
A base RecyclerView.Adapter implementation which takes care of recycling model instances as well as the associated item views.


squidb-addons/squidb-recyclerview/build.gradle
```diff
@@ -0,0 +1,101 @@
...
+            classpath 'com.github.dcendents:android-maven-gradle-plugin:1.3'
+        }
+    }
+}
+
+apply plugin: 'com.android.library'
+apply plugin: 'com.github.dcendents.android-maven'
+
+android {
+    compileSdkVersion 22
```
sbosley: I think this should be 19 to make travis happy.

squidb-addons/squidb-recyclerview/src/com/yahoo/squidb/recyclerview/SquidRecyclerAdapter.java
```diff
@@ -0,0 +1,147 @@
...
+     * Adapter#hasStableIds() hasStableIds()}).
+     *
+     * @param cursor the backing SquidCursor
+     * @param idProperty the column to use for item IDs
+     */
+    public SquidRecyclerAdapter(SquidCursor<M> cursor, Property<Long> idProperty) {
+        this.cursor = cursor;
+        this.idProperty = idProperty;
+        if (idProperty != null) {
+            setHasStableIds(true);
```
sbosley: In SquidCursorAdapter, we say hasStableIds is always true and fall back to using _id to get a row id if the user doesn't specify a different property to use. Should we be consistent across both adapters how we handle this?

squidb-addons/squidb-recyclerview/src/com/yahoo/squidb/recyclerview/SquidRecyclerAdapter.java
```diff
@@ -0,0 +1,147 @@
...
+        if (cursor == null || !cursor.moveToPosition(position)) {
+            throw new IllegalStateException("this should only be called when the cursor is valid");
+        }
+        holder.item.readPropertiesFromCursor(cursor);
+        onBindSquidViewHolder(holder, position);
+    }
+
+    /**
+     * Update the contents of the ViewHolder.itemView to reflect the item at the given position. At this point the
+     * ViewHolder.item is populated with valiues from the backing cursor, so it is not necessary to populate the item
```
sbosley: nit: valiues -> values

squidb-addons/squidb-recyclerview/src/com/yahoo/squidb/recyclerview/SquidRecyclerAdapter.java
```diff
@@ -0,0 +1,147 @@
...
+    public SquidRecyclerAdapter() {
+        this(null, null);
+    }
+
+    /**
+     * Construct a new SquidRecyclerAdapter backed by the specified SquidCursor.
+     *
+     * @param cursor the backing SquidCursor
+     */
+    public SquidRecyclerAdapter(SquidCursor<M> cursor) {
```
sbosley: I think I might vote for removing the cursor parameter from constructors. SquidCursorAdapter doesn't have them, and most cursors will be populated from a loader instead anyways.

sbosley:
Some other things to consider, specifically around trying to work similarly to SquidCursorAdapter:
* Should we provide a LayoutInflater like SquidCursorAdapter does? Alternatively, does SquidCursorAdapter really need to provide a layout inflater?
* Do we need to do anything special for inherited models, or will users just perform the appropriate cast of a ViewHolder subclass in such cases?
* How should we best reconcile the differences between how row ids are handled between the two adapters?

squidb-addons/squidb-recyclerview/src/com/yahoo/squidb/recyclerview/SquidRecyclerAdapter.java
```diff
@@ -0,0 +1,127 @@
...
+
+    /**
+     * Change the SquidCursor backing the adapter, returning the old one. Unlike {@link #changeCursor(SquidCursor)},
+     * the returned old SquidCursor is <em>not</em> closed.
+     *
+     * @param newCursor the new SquidCursor
+     * @return Returns the previously set SquidCursor. If no SquidCursor was previously set, new SquidCursor is the
+     * same instance is the previously set one, null is returned.
+     */
+    public SquidCursor<?> swapCursor(SquidCursor<? extends M> newCursor) {
```
sbosley: Return type should probably be SquidCursor<? extends M> for consistency's sake.

sbosley:
LGTM!

PR #87, opened 2015-09-17
Unit tests for unused

sbosley:
Add unit tests for several core functions that were otherwise unused/untested.

### DIFF
--- a/squidb-tests/src/com/yahoo/squidb/test/TriggerTesterSpec.java
+++ b/squidb-tests/src/com/yahoo/squidb/test/TriggerTesterSpec.java
@@ -16,4 +16,8 @@ public class TriggerTesterSpec {
 
     @ColumnSpec(defaultValue = "0")
     int value2;
+
+    String str1;
+
+    String str2;
 }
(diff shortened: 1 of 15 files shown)


squidb-tests/src/com/yahoo/squidb/test/TriggerTesterSpec.java
```diff
@@ -16,4 +16,8 @@
 
     @ColumnSpec(defaultValue = "0")
     int value2;
+
+    String str1;
+
+    String str2;
```
jdkoren: Want to just make value1 and value2 text columns and use them for all the tests?

jdkoren:
LGTM

squidb-tests/src/com/yahoo/squidb/test/TriggerTesterSpec.java
sbosley: Eh, let's just leave it. We use value1 and value2 for some numeric things in the trigger and I'd rather not update the rest of the test.

PR #88, opened 2015-09-18
Fieldless property generators

sbosley:
Allow creating BasicPropertyGenerators that don't originate from a field in a model spec. This supports a class of plugins that want to add fields to models that don't directly correspond to VariableElements in the model spec.


jdkoren:
LGTM

Issue #44, opened 2015-07-12
SquidCursorLoader isn't compatible with support-v4 Loaders


agatti:
Hi there,

  I'm trying to switch some of my manual SquidCursor management code over to SquidCursorLoader, but the application I'm working with needs Fragment and Activity instances to come from support-v4.

  From what I can see, SquidCursorLoader uses android.content.Loader, which makes it incompatible with android.support.v4.app.Fragment and android.support.v4.app.LoaderManager.  Are there any plans to provide something that can work also for support fragments/activities?

Cheers,
Alessandro

jdkoren:
Hi Alessandro,

At one point we did have a support-v4 compatible loader, but we took it out for two reasons: 

1. We didn't want to have the support library as a dependency
2. Android platform distribution numbers show such a small percentage of devices below API 14 that most apps have stopped supporting those levels, which means they have access to the android.content versions of those classes

That being said, there is an easy workaround for you: copy the source of `SquidCursorLoader` into a new class(perhaps with another name to avoid confusion) and change the import statement to 

```java
import android.support.v4.content.AsyncTaskLoader
```

It should just work, but let me know if you run into other problems with that.

agatti:
Hi Jonathan,

  thank you for your prompt response.  Having my own copy of SquidCursorLoader with that simple change in was indeed what I did and it does seem to work here, albeit I'm a bit worried about future maintenance of the application I'm working on when SquiDB gets updated some time in the future.

  Regarding platform distribution, my application has `minsdk` set to 15, but I am still using the Support Library due to nested fragment support being available only from API 17 and higher.  Unfortunately that's the reason for my application (and I guess quite a few others out there) to use `android.support.v4.app.Fragment` and/or `android.support.v4.ActivityCompat` and thus unable to use `SquidCursorLoader` in its current form.

  Maybe you may want to bring the support library dependency back in until API 17 is the widely accepted `minsdk`?

  Still, thanks for your work, using SquiDB has been a terrific experience so far!

jdkoren:
While you do have a valid point, I think it's likely we will keep things the way they are. We don't want to introduce a dependency in apps that don't have/need it, and the workaround is simple enough; in fact, we already mention it [in our wiki pages](https://github.com/yahoo/squidb/wiki/Support-for-Pre-ICS#porting-squidcursorloader).

agatti:
I understand.  Anyway, thank you for the feedback!

[...]

PR #89, opened 2015-09-21
Support loader

sbosley:
Version of SquidCursorLoader that uses the support library. Refs #44


squidb-addons/squidb-support-loader/src/com/yahoo/squidb/support/SquidCursorLoader.java
```diff
@@ -0,0 +1,116 @@
...
+import com.yahoo.squidb.data.AbstractModel;
+import com.yahoo.squidb.data.SquidCursor;
+import com.yahoo.squidb.data.SquidDatabase;
+import com.yahoo.squidb.sql.Query;
+
+/**
+ * A {@link CursorLoader} that queries a {@link SquidDatabase}. This version of SquidCursorLoader uses the support
+ * library instead of the base Android SDK.
+ */
+public class SquidCursorLoader<T extends AbstractModel> extends AsyncTaskLoader<SquidCursor<T>> {
```
sbosley: @jdkoren do you think this should have the same name as the one in the utility package or should we give it it's own name, like SquidSupportCursorLoader?

jdkoren: I might be in favor of naming it SquidSupportCursorLoader. Granted the Android way is the support library classes have the same name but different package qualifier as the normal ones, but that lends itself to mistakes IMO.

sbosley: Agreed, I'll update it.

jdkoren:
LGTM

PR #91, opened 2015-09-23
Make Criterion.fromRawSelection more forgiving

sbosley:
Log a warning when converting Criterion.fromRawSelection to raw SQL instead of throwing an exception. Document that an exception may be thrown if you try to execute the resulting raw SQL directly.

### DIFF
--- a/squidb/src/com/yahoo/squidb/sql/Criterion.java
+++ b/squidb/src/com/yahoo/squidb/sql/Criterion.java
@@ -6,6 +6,9 @@
 package com.yahoo.squidb.sql;
 
 import android.text.TextUtils;
+import android.util.Log;
+
+import com.yahoo.squidb.utility.VersionCode;
 
 import java.util.Collections;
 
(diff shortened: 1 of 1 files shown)


squidb/src/com/yahoo/squidb/sql/Criterion.java
```diff
@@ -90,10 +100,10 @@ protected void populate(SqlBuilder builder, boolean forSqlValidation) {
                 builder.sql.append(selection);
                 if (selectionArgs != null && selectionArgs.length > 0) {
                     if (builder.args == null) {
-                        throw new UnsupportedOperationException("Raw selection cannot be used in this context--"
-                                + "it cannot converted to raw SQL without bound arguments.");
+                        Log.w("squidb", "Raw selection criterion converted to raw SQL with unbound arguments");
```
jdkoren: Is it worth having a global log tag for all our log calls?

jdkoren:
LGTM, one comment we can address separately if desired

squidb/src/com/yahoo/squidb/sql/Criterion.java
sbosley: Probably yes. I'll put it in the next PR I'm about to open.

PR #96, opened 2015-10-08
Misc enhancements

sbosley:
A bunch of minor changes for better consistency. Makes sure SquidDatabase wraps insert/insertOrThrow correctly and distinguishes between the two, and unifies some duplicated logic in SqlUtils.

ThreadLocals are now accepted as arguments to criterions, which is a step towards better reuse of statements in a multi-threaded environment.

Also adds the ability to construct ConjunctionCriterions from Lists.

### DIFF
--- a/squidb/src/com/yahoo/squidb/sql/SqlUtils.java
+++ b/squidb/src/com/yahoo/squidb/sql/SqlUtils.java
@@ -7,7 +7,6 @@ package com.yahoo.squidb.sql;
 
 import android.text.TextUtils;
 
-import java.math.BigDecimal;
 import java.util.Collection;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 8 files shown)


squidb/src/com/yahoo/squidb/sql/SqlUtils.java
```diff
@@ -7,7 +7,6 @@
 
 import android.text.TextUtils;
 
-import java.math.BigDecimal;
```
jdkoren: Is this no longer supported? Or do we just no longer need to check for it specifically?

sbosley: It was handled inconsistently. We'd inline it as a String in decimal form, but DatabaseUtils would bind it as a Long (since it subclasses Number). I think it's better to stick with the DatabaseUtils behavior for consistency, unless we want to reimplement DatabaseUtils ourself. Since we'd just coerce to a potentially-lower-precision double anyways though, it makes more sense for users to use Double or AtomicReference<Double>

jdkoren:
LGTM, awaiting travis

PR #103, opened 2015-10-22
Function.substr enhancements

sbosley:
Make Function.substr work with arbitrary arguments instead of hardcoded int literals


jdkoren:
LGTM

PR #104, opened 2015-10-23
WAL test

sbosley:
I added a test case that ensures concurrent reads with WAL were always working, and I discovered that such a test would fail on API 15 because we enable WAL in onConfigure, which was added in API 16. I thought we could patch this at the open helper level by having the open helper for stock android manually call onConfigure after onOpen. @jdkoren what do you think? It seems like a nice convenience for older API levels still be able to use the onConfigure hook. The order of callbacks will be slightly different pre-16, but I doubt that would matter much for most use cases.

### DIFF
--- a/squidb/src/com/yahoo/squidb/data/adapter/DefaultOpenHelperWrapper.java
+++ b/squidb/src/com/yahoo/squidb/data/adapter/DefaultOpenHelperWrapper.java
@@ -55,10 +55,10 @@ public class DefaultOpenHelperWrapper extends SQLiteOpenHelper implements SQLite
     @Override
     public void onOpen(SQLiteDatabase db) {
         SQLiteDatabaseAdapter adapter = new SQLiteDatabaseAdapter(db);
-        delegate.onOpen(adapter);
         if (Build.VERSION.SDK_INT < Build.VERSION_CODES.JELLY_BEAN) {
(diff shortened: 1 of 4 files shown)


squidb/src/com/yahoo/squidb/data/adapter/DefaultOpenHelperWrapper.java
```diff
@@ -53,7 +54,11 @@ public void onConfigure(SQLiteDatabase db) {
 
     @Override
     public void onOpen(SQLiteDatabase db) {
-        delegate.onOpen(new SQLiteDatabaseAdapter(db));
+        SQLiteDatabaseAdapter adapter = new SQLiteDatabaseAdapter(db);
+        delegate.onOpen(adapter);
```
jdkoren: Since `onConfigure()` comes before `onOpen()` in AOSP, maybe we should keep that ordering and move this call down.

sbosley: Well, onConfigure comes before EVERYTHING in AOSP, so this will be a different ordering no matter what, but I suppose we could at least move it before onOpen, sure.

jdkoren:
LGTM

PR #107, opened 2015-10-28
SQLite bindings version bump

sbosley:
This PR bumps the SQLite version in the SQLite bindings project to 3.9.1, but also removes the SQLite source files from version control. They only need to be there when building from source, so instead I just left a note in a readme file noting that anyone building from source should add the amalgamation files to the jni/sqlite directory.

Note that pre-built native binaries for the bindings project are left in so that travis can test against them.


jdkoren:
LGTM

PR #111, opened 2015-11-03
Add the @Constants annotation for constant holding static inner classes in

sbosley:
Most constants can safely be copied directly to the generated model. However, constants that are defined using table schema fields (e.g. from Properties, Table, View, like public static final Order DEFAULT_ORDER = Model.COLUMN.asc()) may run into class loading order issues where the fields needed to initialize the constant have not yet themselves been initialized at class loading time. (This is discussed in more detail in #110.

This commit allows constants to be declared in a static inner class of the model spec, effectively deferring initialization of copied constants until after the full schema has been initialized. The old behavior of directly copying constants has been left intact so those who don't need this control over class loading order can use things as they always have.


jdkoren:
LGTM

Issue #112, opened 2015-11-04
QueryBuilder with join - reading multiple models with same fields (ID)


MFlisar:
Here's how I tried to use it, but this had one problem: the default ID is wrong. The MediaTag reads the ID from the Tag model...

	Query query = Query.select().from(MediaTag.TABLE).leftJoin(Tag.TABLE, MediaTag.FK_TAG.eq(Tag.ID));
	SquidCursor<MediaTag> cursor = MainApp.getDao().query(null, query);
	cursor.moveToFirst();
	while (!cursor.isAfterLast())
	{
		MediaTag mediaTag = new MediaTag();
		Tag tag = new Tag();
		mediaTag.readPropertiesFromCursor(cursor);
		tag.readPropertiesFromCursor(cursor);
		tags.add(new WrappedTag(tag, mediaTag));
		cursor.moveToNext();
	}
	cursor.close();

This happens because the query gives all fields simple aliases without table prefixes. Do I really have to manually create aliases and use them and read them? Following works:

	List<Field<?>> fields = new ArrayList<>();
	fields.add(MediaTag.ID.as("ID1"));
	... other fields ...
	fields.add(Tag.ID.as("ID2"));
	... other fields ...

	Query query = Query.select(fields).from(MediaTag.TABLE).leftJoin(Tag.TABLE, Tag.ID.eq(MediaTag.FK_TAG)).orderBy(MediaTag.PATH.asc());
	SquidCursor<MediaTag> cursor = MainApp.getDao().query(null, query);
	cursor.moveToFirst();
	while (!cursor.isAfterLast())
	{
		MediaTag mediaTag = new MediaTag();
		Tag tag = new Tag();
		mediaTag.readPropertiesFromCursor(cursor);
		tag.readPropertiesFromCursor(cursor);
		tag.setId(cursor.getLong(3));
		mediaTag.setId(cursor.getLong(0));
		tags.add(new WrappedTag(tag, mediaTag));
		cursor.moveToNext();
	}
	cursor.close();

Is there an easier way? That does not require to manually define fields? Or can I somehow enable to use no aliases in the query and use prefixed fields instead?

sbosley:
At the moment, you will need to explicitly give columns with the same name aliases for the query.

[...]

PR #113, opened 2015-11-10
qualifyField fix

sbosley:
Came across a couple issues with aliasing while investigating potential enhancements for #112. Not ready to commit anything for that issue just yet but I think we'll want these fixes regardless.

### DIFF
--- a/squidb/src/com/yahoo/squidb/sql/Query.java
+++ b/squidb/src/com/yahoo/squidb/sql/Query.java
@@ -6,7 +6,6 @@
 package com.yahoo.squidb.sql;
 
 import com.yahoo.squidb.data.ViewModel;
-import com.yahoo.squidb.sql.Property.LongProperty;
 import com.yahoo.squidb.utility.SquidUtilities;
 
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


squidb/src/com/yahoo/squidb/sql/Query.java
```diff
@@ -496,18 +495,15 @@ private void visitSelectClause(SqlBuilder builder, boolean forSqlValidation) {
             builder.sql.append("DISTINCT ");
         }
 
+        List<Field<?>> toSelect;
         if (isEmpty(fields)) {
-            // Explicitly add rowid for virtual tables: "select table.rowid as rowid, *"
-            if (table instanceof VirtualTable) {
```
jdkoren: What was this for originally, and is it safe to remove it?

sbosley: It was originally because "select *" doesn't include columns named rowid. It's safe to remove because now we explicitly inline the rowid field rather than using *. Also there should be a unit test making sure that fetching rowids in the select all case works for virtual table models.

jdkoren:
LGTM

Issue #116, opened 2015-11-23
Negating an "In" clause


matthew-reilly:
How would one construct a query like the following? Doesn't seem to be currently possible. 

```
SELECT *
FROM employees
WHERE department NOT IN ('Accounting', 'Finance', 'HR');
```

sbosley:
Easy! One of two ways:

```java
Criterion.not(Employee.DEPARTMENT.in("Accounting", "Finance", "HR"));
// or...
Employee.DEPARTMENT.in("Accounting", "Finance", "HR").negate();
```
Hope that helps!

matthew-reilly:
Ah, stupid simple. Gracias. :shipit:

sbosley:
No problem, happy to help! In case you haven't seen them yet, for future reference most of the common use cases for the query builder are documented on [this wiki page](https://github.com/yahoo/squidb/wiki/SquiDB's-query-builder), or for more advanced things you can check out [this page](https://github.com/yahoo/squidb/wiki/Advanced-Query-usage). They may not cover 100% of the possible capabilities of Query (which is pretty extensive), but they're a good reference for most use cases.

matthew-reilly:
It would see that generates a query that looks like this:

```
( NOT (employee.department IN ('Accounting','Finance','HR')))
```

Which, from what I know about sqlite, isn't valid sql? Shouldn't it be NOT IN?

sbosley:
"NOT IN" is indeed acceptable SQL, but the form it generates is also valid and equivalent -- SQLite describes "NOT" as a valid unary prefix operator on [this page](https://www.sqlite.org/lang_expr.html).

I *think* that SQLite will always choose equivalent query plans for these two forms. We can consider an enhancement to make the generated SQL more readable though, for debugging purposes, and/or add a `notIn(...)` method to Field as a counterpart to `in(...)`.

matthew-reilly:
Ah, will trust you then! Will continue my testing. 

A notIn would be helpful, perhaps if I have some extra time I'll make a PR. 

Cheers.

PR #117, opened 2015-11-23
Criterion negation enhancements

sbosley:
Add explicit negation methods for many of the standard criterions. E.g. `in(...)` now has a `notIn(...)` counterpart, `like(...)` has `notLike(...)`, etc. Each of these criterions will also generate more readable SQL when negated. Refs #116.


jdkoren:
LGTM

PR #120, opened 2015-12-04
Pretty print cursor

sbosley:
Add a utility for printing a formatted cursor. This will be pretty close to what you see in the SQLite console with .headers on and .mode columns, and should be much more readable than the results of DatabaseUtils.dumpCursor().


jdkoren:
Code LGTM

PR #124, opened 2015-12-18
Allow using SQL expressions for limit and offset in Query

sbosley:
e.g. for when you want to use something like the result of a scalar subquery in limit/offset


jdkoren:
LGTM

PR #126, opened 2016-01-07
Make it easy to copy/generate javadocs in generated model classes (3.0 only)

sbosley:
Add a JavadocPlugin that can copy javadocs from fields/methods/model spec classes to their corresponding entities in the generated class.

This is against the 3.0 branch because we use some of the updated plugin APIs.

### DIFF
--- a/squidb-processor/src/com/yahoo/squidb/processor/writers/ModelFileWriter.java
+++ b/squidb-processor/src/com/yahoo/squidb/processor/writers/ModelFileWriter.java
@@ -80,6 +80,7 @@ public abstract class ModelFileWriter<T extends ModelSpec<?>> {
 
         emitPackage();
         emitImports();
+        plugins.beforeEmitClassDeclaration(writer);
         beginClassDeclaration();
 
         plugins.beforeEmitSchema(writer);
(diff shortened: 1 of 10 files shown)


squidb-tests/src/com/yahoo/squidb/test/TestModelSpec.java
```diff
@@ -81,6 +100,13 @@ public static void run(TestModel instance) {
         return null;
     }
 
+    /**
+     * It would be pretty great if this static method had a javadoc too
```
jdkoren: LOL

squidb-processor/src/com/yahoo/squidb/processor/writers/ModelFileWriter.java
```diff
@@ -115,8 +116,6 @@ private void emitImports() throws IOException {
     }
 
     private void beginClassDeclaration() throws IOException {
-        writer.writeComment("Generated code -- do not modify!");
-        writer.writeComment("This class was generated from the model spec at " + modelSpec.getModelSpecName());
```
jdkoren: It might be nice to still have these lines somewhere as it's a good warning to someone who opens the file

sbosley: Agreed! That's why they've been moved to the new JavadocPlugin :)

jdkoren:
LGTM

PR #128, opened 2016-01-09
Rewritten JSON plugin (3.0)

sbosley:
Our old Jackson plugin has never been much more than a fun experiment from the early days of the code generation plugin system. With the introduction of [the json1 extension](http://sqlite.org/json1.html) in SQLite 3.9.0, it made sense to rewrite that plugin to be a significantly more powerful extension that can take advantage of the new features when connecting to SQLite builds that support them.

This rewrite comes with two major feature additions:
 * The `squidb-json-compiler` module now supports properties annotated with `@JSONField` for serializing arbitrary objects (POJOs, Lists, Maps, etc.) to a JSON string and storing that string in a column. This requires providing a JSON serialization engine implementing the simple new `JSONMapper`. The major advantage besides being able to support arbitrary objects is that the serialization is library-agnostic; any JSON library can be used to support the serialization/deserialization. Two simple examples of such implementations can be found in the `JSONTestCase` class, using Jackson and Gson as the libraries for serialization.
 * The `squidb-json-plugin` runtime module adds support for the json1 functions for examining/manipulating JSON strings directly from SQL. These functions are declared in a `JSONFunctions` class, similar to the existing SquiDB `Functions` class for built-in SQL functions. Obviously these functions will only work if connecting to a SQLite build with the json1 extension enabled, so the squidb-sqlite-bindings project now enables this extension in its precompiled SQLite build by default.

All modules in the squidb-json addon are now plain jars with no Android dependencies. We only test on Android at the moment, but theoretically there's no reason why this couldn't also work on iOS under j2objc given an iOS-compatible implementation of `JSONMapper`.


squidb-addons/squidb-json/squidb-json-plugin/src/com/yahoo/squidb/json/JSONPropertySupport.java
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Business logic of managing serialization and deserialization of JSON properties. Clients of this plugin should
+ * be sure to call {@link #setJSONMapper(JSONMapper)} to initialize the serialization engine. See the JSONPropertyTest
+ * class in the squidb tests project for example implementations of the JSONMapper interface.
+ */
+public class JSONPropertySupport {
+
+    private static final String TAG = "squidb-json";
+
+    public static JSONMapper MAPPER;
```
jdkoren: Probably should be private so we could add some kind of defensive code in `setJSONMapper()`

jdkoren:
LGTM

PR #130, opened 2016-01-23
Enum support

jdkoren:


squidb-processor/src/com/yahoo/squidb/processor/plugins/defaults/properties/EnumFieldReferencePlugin.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Override
+    public boolean hasChangesForModelSpec() {
+        return modelSpec instanceof ViewModelSpecWrapper || modelSpec instanceof InheritedModelSpecWrapper;
+    }
+
+    @Override
+    protected boolean hasPropertyGeneratorForField(VariableElement field, DeclaredTypeName fieldType) {
+        return field.getModifiers().containsAll(TypeConstants.PUBLIC_STATIC_FINAL)
+                && fieldType.equals(TypeConstants.ENUM_PROPERTY);
```
sbosley: nit: let's do TypeConstants.ENUM_PROPERTY.equals(fieldType)

squidb-processor/src/com/yahoo/squidb/processor/plugins/defaults/properties/EnumFieldReferencePlugin.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Override
+    protected PropertyGenerator getPropertyGenerator(VariableElement field, DeclaredTypeName fieldType) {
+        // We know it's an EnumProperty, so extract the type arg
+        List<? extends TypeName> typeArgs = fieldType.getTypeArgs();
+        if (typeArgs != null && typeArgs.size() == 1) {
+            TypeName enumType = typeArgs.get(0);
+            if (enumType instanceof DeclaredTypeName) {
+                return new EnumPropertyGenerator(modelSpec, field, utils, (DeclaredTypeName) enumType);
+            }
```
sbosley: Can we log a warning in the else case?

jdkoren: Sure. Do you prefer a warning or an error?

It's worth noting the only way I was able to hit this case was to write a Model class by hand which declared a raw `EnumProperty`, then make a `@ViewModelSpec`-annotated class that referenced it (also with a raw `EnumProperty`). We might want to discuss offline a little more, as one of my attempts actually resulted in a StackOverflowError in AptUtils.

sbosley: I think a warning is fine. It'll just be ignored if it doesn't fit the criteria.

I was assuming you could also hit it by having a regular EnumProperty in a table model but writing it in the view model spec like so:
```java
// This is no good, the type arg needs to be explicit
public static final EnumProperty<?> ENUM = Model.SOME_ENUM;
```

StackOverflowError is no good though -- let's definitely look into that. Maybe separately though.

sbosley: On second thought, I think maybe an error does make more sense -- we might want to fail the build early so that users can track down why their enum properties aren't being added to the view model.

jdkoren: Whoops, saw the first message and pushed the change. Can still make it an error (although I notice that with an error, gradle won't show the actual error, just compile errors for not finding symbols because various models weren't generated, whereas with warning it will).

sbosley: Annoying. Warning is fine then.

jdkoren: At least if they haven't disabled getters & setters, it won't generate those properly and there will likely be other compile errors to stop them (as is the case with our test suite).

sbosley: Won't it just not generate them at all, since it's skipped adding a PropertyGenerator for that property?

sbosley:
LGTM, nice work! Anything else you can think of we need to on this one or is it good to go?

squidb-processor/src/com/yahoo/squidb/processor/plugins/defaults/properties/EnumFieldReferencePlugin.java
jdkoren: Yeah, I got confused because I saw it copied with the other constants like this:

```java
// --- constants
public static final EnumProperty<?> TEST_ENUM = TestViewModelSpec.TEST_ENUM;
```

jdkoren:
Probably good to go. We already discussed that cursors don't support it yet, other than that I can't think of anything.

sbosley:
:+1: Merging!

PR #132, opened 2016-01-28
iOS custom sqlite builds + json functions (3.0)

sbosley:
Two related changes here:
1) Update our iOS test script to verify that things still work as expected if SQLite is built from scratch instead of using the -lsqlite3 linker flag.
2) Move our JSON tests out of the android test package so that they can run on iOS given a compatible SQLite build.


jdkoren:
LGTM

PR #133, opened 2016-02-02
Locking enhancements

sbosley:
This PR allows methods like close(), clear(), and recreate() to be called more safely in a multi-threaded environment. Calling these methods now acquires the exclusive lock, so that any ongoing transactions can finish before the DB is closed. (Open cursors may still abruptly be closed, however).


jdkoren:
LGTM

PR #134, opened 2016-02-03
More minor locking tweaks

sbosley:
Minor tweaks to be slightly more defensive when we acquire the non-exclusive lock for transactions. If anything goes wrong, we should release the lock before we throw the exception, so that users who are suppressing exceptions at least won't have any chance of deadlock.


jdkoren:
LGTM

PR #135, opened 2016-02-04
More exception handling improvements

sbosley:
This PR makes SquidDatabase more disciplined about the cleanup we do if runtime exceptions are encountered while opening a DB -- most importantly, we should null out the internal database instance to make sure there's no way we could accidentally be holding an invalid one in case the thrown exception is silently suppressed.

We also should stop wrapping runtime exceptions in other exceptions -- it makes things more unpredictable if you're trying to catch runtime problems and doesn't really have any benefit.

I'll add a few unit tests before we merge this.


jdkoren:
LGTM

PR #137, opened 2016-02-07
Database instance lock

sbosley:
Protect the internal database instance with a private lock, rather than synchronizing on this. This is functionally the same, but protects users in the case that they create any synchronized methods that access the DB in their SquidDatabase subclass.


jdkoren:
LGTM

PR #139, opened 2016-02-17
Fix doclint errors when using java8

smb510:
When you compile Squidb from source using java 8, the javadoc task uses a really strict new doclint, which fails when non-HTML escaped greater than and less than characters are used (since it interprets them as HTML tags). This PR replaced them with proper HTML so that javadocs are generated properly with the strict doclint.

Weird, huh?


sbosley:
Thanks @smb510! Note too that the SquiDB project itself works around this issue this way: https://github.com/yahoo/squidb/blob/master/build.gradle#L22

smb510:
@sbosley oh, that must not be in the tag I'm using... i did actually try that :)

smb510:
there's some other places where this won't work. I'm going to close for now, and reopen a new PR in a short while

PR #140, opened 2016-02-17
fix java8 doclint errors that prevent compilation

smb510:
In java8, doclint does not allow the use of "<" or ">" characters, and will fail compilation. This PR replaces the instances of those characters with HTML-friendly ones.


smb510:
This PR supersedes #139

sbosley:
:+1:

jdkoren:
LGTM

PR #141, opened 2016-02-23
Open DB error handling

sbosley:
This PR adds an optional, overridable hook in SquidDatabase that allows the user to attempt to recover if the database fails to open for any reason (i.e., if we catch a RuntimeException trying to open the DB). There aren't too many actions that can be taken here, since the DB isn't open, so the hook is currently designed largely around the actions that are reasonable -- attempting to reopen the DB with a recursive call to `getDatabase()`, or deleting and recreating the DB using `recreate()`. A `retryCount` parameter incremented on each call to the hook allows clients to abort/crash instead of getting caught in an infinite loop if even deleting and recreating the DB file doesn't fix the problem. By default, if this hook is not overridden, it will simply rethrow whatever exception was forwarded to it (meaning if you don't use the new hook, behavior is exactly the same as it was before).

In future versions, we should consider if a checked exception approach would be better, but that would a) be a backwards-incompatible change and b) break from the Android pattern of preferring to throw RuntimeException subclasses for SQLite problems, so for now this seems like the safest option.

This PR is tagged with WIP so we can continue to consider the API and gather a bit of feedback before we merge it.

### DIFF
--- a/squidb-tests/src/com/yahoo/squidb/data/SquidDatabaseTest.java
+++ b/squidb-tests/src/com/yahoo/squidb/data/SquidDatabaseTest.java
@@ -111,7 +111,19 @@ public class SquidDatabaseTest extends DatabaseTestCase {
         // close and reopen to trigger an upgrade/downgrade
         badDatabase.onTablesCreatedCalled = false;
         badDatabase.close();
-        badDatabase.getDatabase();
+        RuntimeException caughtException = null;
+        try {
+            badDatabase.getDatabase();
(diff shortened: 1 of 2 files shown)


sbosley:
Let's give this another thorough review, maybe add one or two more test cases, and then get it merged. We can leave it tagged with `@Beta` to indicate that it's less developed than other APIs and may change in the future.

jdkoren:
Let's merge so we can start testing it

PR #142, opened 2016-03-11
JSONMapper type argument

sbosley:
Make JSONMapper.toJSON take a Type as an argument so implementations can better handle generic types


jdkoren:
LGTM

PR #145, opened 2016-03-17
Fix strings on iOS

sbosley:
We had a bug in our iOS port of CursorWindow where we were not correctly converting strings back to UTF-16 the way the native Android implementation does. This PR borrows some more code from AOSP to more exactly mimic the Android behavior (code found in SquidbUnicode.h and SquidbUnicode.m, slightly adapted for our use case).


sbosley:
I should write a couple unit tests before we merge this.

jdkoren:
LGTM

PR #149, opened 2016-03-21
Version bumps

sbosley:
This PR bumps various dependency versions and addresses minor cleanup tasks:
* Bump android gradle plugin version used by all modules to 1.5.0
* Bump compileSdkVersion to 23 and build tools to 23.0.2
* Add android-23 as a target for our travis builds
* Bump the SQLite version in squidb-sqlite-bindings to 3.11.1
* Remove all the old .properties files that aren't needed by the gradle build system

Travis was having trouble running on the android-23 emulators due to memory limitations, but they do pass on my machine.


jdkoren:
LGTM

PR #150, opened 2016-03-21
mapToModel fixes

sbosley:
ViewModel.mapToModel would break down when the view model was composed of multiple aliased joins on the same table. This PR adds a new version of the mapToModel method that takes an aliased table argument to correct this problem. The single-arg version will still work as expected if the view model joins on a table only once, but will throw an exception to indicate programmer error if it detects multiple aliases for the same class are present.

### DIFF
--- a/squidb/src/com/yahoo/squidb/data/ViewModel.java
+++ b/squidb/src/com/yahoo/squidb/data/ViewModel.java
@@ -5,12 +5,15 @@
  */
 package com.yahoo.squidb.data;
 
+import android.util.Log;
+
 import com.yahoo.squidb.sql.Property;
 import com.yahoo.squidb.sql.Property.PropertyWritingVisitor;
 import com.yahoo.squidb.sql.Query;
 import com.yahoo.squidb.sql.SqlTable;
 import com.yahoo.squidb.sql.Table;
 import com.yahoo.squidb.sql.View;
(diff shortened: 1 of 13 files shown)


squidb/src/com/yahoo/squidb/data/ViewModel.java
```diff
@@ -184,20 +210,46 @@ protected static void validateAliasedProperties(Property<?>[] aliasedPropertyArr
...
+                return null;
+            }
+            if (table == null) {
+                if (visitors.size() == 1) {
+                    for (TableModelMappingVisitor<?> visitor : visitors.values()) {
+                        return (TableModelMappingVisitor<T>) visitor;
+                    }
+                } else {
+                    // Log an error and return null
+                    Log.e(SquidUtilities.LOG_TAG, "Attempted to mapToModel for class " + cls + ", but multiple" +
```
sbosley: Open question -- should this throw an exception instead of returning null? Might be better for tracking down usage errors.

sbosley:
I'd like to also address #152 with this PR, so let's hold off on merging it until I can get a handle on that as well.

sbosley:
I've added the changes that fix #152, so this is now ready for a good thorough review.

jdkoren:
LGTM

PR #153, opened 2016-03-28
Fix handling of static final fields in model specs

sbosley:
There was a bug in the code generator that would create column definitions from private static final fields in model specs, since it only considered public static final fields to be constants. This fixes the bug, and also allows the constant copying plugin to copy any package-level or higher static final fields as constants to the generated model class.


jdkoren:
LGTM

PR #159, opened 2016-04-12
Disable data change notifiers during db open/upgrade

### LINKED ISSUE #154: Doing DB writes during open/upgrade hooks can trigger data change notifiers
In many cases this wouldn't be a problem, but it could be in some use cases and it doesn't really make sense for data change notifiers to be watching before the db has even opened anyways. Clients can work around any issues right now by using setDataChangedNotificationsEnabled during these hooks, but we should consider how to best handle this at the SquidDatabase level too.

sbosley:
Closes #154


jdkoren:
LGTM

PR #162, opened 2016-04-14
Fix Java8 doclint errors

beparker:
Java8 doclint requires escaped HTML characters.  This PR also fixes an invalid "@see" reference to onMigrationFailed in SquidDatabase.java.

### DIFF
--- a/squidb/src/com/yahoo/squidb/data/SquidDatabase.java
+++ b/squidb/src/com/yahoo/squidb/data/SquidDatabase.java
@@ -158,9 +158,9 @@ public abstract class SquidDatabase {
      * switch(oldVersion) {
      * boolean result = true;
      * case 1:
-     *     result &= tryAddColumn(MyModel.NEW_COL_1);
+     *     result &amp;= tryAddColumn(MyModel.NEW_COL_1);
      * case 2:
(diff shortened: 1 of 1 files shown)


squidb/src/com/yahoo/squidb/data/SquidDatabase.java
```diff
@@ -175,7 +175,7 @@ protected void onTablesCreated(SQLiteDatabaseWrapper db) {
      * @param oldVersion the current database version
      * @param newVersion the database version being upgraded to
      * @return true if the upgrade was handled successfully, false otherwise
-     * @see {@link #onMigrationFailed(MigrationFailedException)}
+     * @see onMigrationFailed(MigrationFailedException)
```
sbosley: When I do this Android Studio gives me an error. What was the issue you were having with it?

sbosley: I think maybe it should be `@see #onMigrationFailed`: http://docs.oracle.com/javase/7/docs/technotes/tools/windows/javadoc.html#see

beparker: Fixed in b7504b1

sbosley:
LGTM

PR #164, opened 2016-04-18
Generate javadoc link back to model spec in generated model class (#163)

### LINKED ISSUE #163: Turn single-line comment into doc comment to allow effortless navigation from a table model to its spec
At `com.yahoo.squidb.processor.writers.ModelFileWriter.beginClassDeclaration(ModelFileWriter.java:135)`, I suggest writing a doc comment instead of a single line comment and adding a Javadoc `@link`.

So, instead of the following:

```Java
// Generated code -- do not modify!
// This class was generated from the model spec at x.y.z.MyTableModel
```

Consider this:

```Java
// Generated code -- do not modify!
/** This class was generated from the model spec at {@link x.y.z.MyTableModel} */
```

At least that would make the link clickable in Android Studio. I'm not sure whether that would render a broken link in any generated Javadoc, however.

sbosley:
Closes #163


jdkoren:
LGTM

PR #166, opened 2016-04-21
Fixes #165, ModelSpecProcessor handles bad input gracefully

### LINKED ISSUE #165: Annotation processor fails with unexpected input (Kotlin)
The `ModelSpecProcessor` will attempt to process all annotations it receives, even if they're part of the list of supported annotations returned by `getSupportedAnnotationTypes()`.

From what I could tell, the JSR-269 specification for annotation processing doesn't explicitly say that a processor will only receive annotations it supports, although it is implied.

The specific use case is with Kotlin and its own processing tool (kapt), there are some workarounds that appear to result in the processor being handed annotations outside of its own supported ones. By skipping these other annotations in `ModelSpecProcessor`, SquiDB works with Kotlin as well based on my testing.

I plan on filing an issue with Kotlin as well, since that doesn't seem like desired behavior, but it's pretty simple to work around as well. I can create a pull request if you feel like this change is desirable.

arvola:
Makes ModelSpecProcessor skip processing annotations it doesn't support and log a warning when that happens.

### DIFF
--- a/squidb-processor/src/com/yahoo/squidb/processor/ModelSpecProcessor.java
+++ b/squidb-processor/src/com/yahoo/squidb/processor/ModelSpecProcessor.java
@@ -54,18 +54,22 @@ import javax.tools.Diagnostic.Kind;
 @SupportedSourceVersion(SourceVersion.RELEASE_6)
 public final class ModelSpecProcessor extends AbstractProcessor {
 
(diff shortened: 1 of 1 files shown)


jdkoren:
LGTM

sbosley:
LGTM too. Thanks for this @arvola!

PR #167, opened 2016-04-22
Fix spurious code gen warning

sbosley:
Don't need to warn about private static final fields in model specs since they'd generally be internal constants.


jdkoren:
LGTM

PR #168, opened 2016-04-23
Misc cleanup

sbosley:
Delete some unused junk and clean up our codebase to use the diamond operator and fix a bunch of harmless warnings.


squidb-tests/src/com/yahoo/squidb/sql/InsertTest.java
```diff
@@ -273,7 +273,7 @@ public void run() {
                 // insert into testModels (firstName, lastName) values ("Jack", "Sparrow"), ("James", "Bond", 007),
                 // ("Bugs", "Bunny");
                 Object[] values1 = new Object[]{"Jack", "Sparrow"};
-                Object[] values2 = new Object[]{"James", "Bond", Integer.valueOf(007)};
```
jdkoren: aw...

sbosley: Haha sorry. I was cleaning up the "unnecessary boxing" warnings and didn't notice the actual context here.

squidb/src/com/yahoo/squidb/utility/SquidUtilities.java
```diff
@@ -186,7 +186,7 @@ private static void addColumnToRowBuilder(StringBuilder builder, String value, i
      * A version of {@link Collection#addAll(Collection)} that works on varargs without calling Arrays.asList, which is
      * a performance and memory boost
      */
-    public static <T> void addAll(Collection<T> collection, T... objects) {
+    public static <T> void addAll(Collection<T> collection, T[] objects) {
```
jdkoren: i thought this was intended to be varargs?

sbosley: It's really intended as a wrapper to Collections.addAll, but it does a null check. It's never called directly as a varargs method itself, but instead callers pass their own varargs argument or an array, whichever they happen to have.

jdkoren:
LGTM

PR #169, opened 2016-04-29
Fix memory leaks (3.0)

sbosley:
On our 3.0 branch, we had a memory leak (iOS only) where a strong reference cycle between a SqlTable and its properties could be created, since they both held references to each other. Properties don't actually need the entirety of the table object though -- they only need a table name and a model class. To break the reference cycle, this PR introduces `TableModelName`, which is a tuple of model class and table name. Property objects are now constructed with and hold a reference to this tuple in lieu of a SqlTable object.


jdkoren:
LGTM

PR #170, opened 2016-05-09
iOS cursor window native dispose

sbosley:
Our iOS port of CursorWindow had been relying on the deallocation of CursorWindowNative to free the large buffer allocated for holding cursor window data. While this would technically work eventually, it can leave the buffer hanging around for a long while if the cursor window is autoreleased but not deallocated immediately. The Android behavior is to dispose of the large native buffer as soon as the window is closed, and then allow garbage collection to collect the smaller JVM object. We should mirror this behavior on iOS by immediately freeing the native cursor window buffer when it can be disposed of using the already-existing nativeDispose hook.

Holding for additional testing but this should be ready to go soon.


jdkoren:
Looks reasonable. Let me know how testing goes and I'll give it a LGTM.

sbosley:
Testing looks solid.

jdkoren:
LGTM

PR #171, opened 2016-05-13
Fix harmless warnings emitted by apt-utils in squidb-processor

sbosley:
Update squidb-processor to use java 1.7 and the latest version of apt-utils, which removes several unecessary warnings


jdkoren:
💎 s are forever.

LGTM

PR #175, opened 2016-06-03
Move plugins to root

jaredsburrows:
@sbosley Moved plugins to root to remove duplicated code

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,11 +7,13 @@ buildscript {
     repositories {
         jcenter()
     }
+
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.0.0'
+        classpath 'com.android.tools.build:gradle:2.1.0'
+        classpath 'com.neenbedankt.gradle.plugins:android-apt:1.8'
+        classpath 'com.github.dcendents:android-maven-gradle-plugin:1.3'
     }
 }
-
 
 allprojects {
     repositories {
--- a/squidb-addons/squidb-jackson-annotations/build.gradle
+++ b/squidb-addons/squidb-jackson-annotations/build.gradle
@@ -10,8 +10,8 @@ version = '2.0.3'
 apply plugin: 'java'
 apply plugin: 'maven'
 
-sourceCompatibility = 1.7
-targetCompatibility = 1.7
+sourceCompatibility = JavaVersion.VERSION_1_7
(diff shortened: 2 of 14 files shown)


jdkoren:
@sbosley Is anything adversely affected by removing the gradle version check? I don't recall what it was for.

jaredsburrows:
@jdkoren Yes, it used an older plugin for older gradle versions. Since we should all be using the wrapper and the latest plugin, I just removed the old one.

sbosley:
@jaredsburrows it's hypothetically possible that there are other users of this project who have not yet updated their gradle version and are building from source. Since this would technically be a backwards-incompatible change for such people, we might want to keep the version check on master and only remove it on our dev_3.0 branch instead.

sbosley:
@jaredsburrows hmm I'm pretty sure that we had a problem once where this would break depending on the version of gradlew used by the root project (since I think the root project gradle is the one that gets used for building, not whatever version in the source dependencies happen to have). Now that I'm thinking of it, have you confirmed that this PR doesn't cause any problems when using squidb as a source dependency in some other project? We should probably test that before merging.

jaredsburrows:
@sbosley Why would you ever use as a source dependency? That defeats the entire purpose of the `android-maven-gradle-plugin` plugin. All you have to do is use `gradlew install` when making changes.

sbosley:
@jaredsburrows the `android-maven-gradle-plugin` is for us; it supports building the artifacts that we deploy to jcenter. However, some projects import squidb as a dependency from source (either for easier debugging, because they're using a slightly altered fork or a non-master branch, or for any number of other reasons). For these projects, the maven plugin does nothing useful, but it's still there and needs to be resolved since it's a part of the source code. I still think we need to test to make sure this change doesn't break projects for people using squidb in that way, and I'm not sure that I understand your objection -- do you see some way of removing that plugin for people building from source without breaking our ability to build artifacts for jcenter?

jaredsburrows:
@sbosley Don't remove the plugin. It is great for publishing to mavenLocal() for testing. People should not be importing the library from source. They should always be using binary dependencies.

sbosley:
Since the gradle version check is for a very niche use case that's *probably* not relevant to anyone anymore, I think after much discussion that it is probably ok to remove. If anyone complains, we can add it back in. @jaredsburrows thanks for taking the time to do the cleanup here!

PR #176, opened 2016-06-03
Move versions to root build.gradle

jaredsburrows:
@sbosley Moving duplicated version to root.

version and group need to only be specified once at the root and `project.ext.artifactId` is not used.


sbosley:
👍 awesome!

PR #177, opened 2016-06-03
Remove extra gitignores

jaredsburrows:
@sbosley Single gitignore at the root


sbosley:
Good catch. Wonder why those were even still in here, good to be rid of them.

PR #178, opened 2016-06-03
Add bintray badge

jaredsburrows:
@sbosley Add bintray's badge

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
-[![Build Status](https://travis-ci.org/yahoo/squidb.svg?branch=master)](https://travis-ci.org/yahoo/squidb)
+[![Build Status](https://travis-ci.org/yahoo/squidb.svg?branch=master)](https://travis-ci.org/yahoo/squidb) [![Download](https://api.bintray.com/packages/yahoo/maven/squidb/images/download.svg) ](https://bintray.com/yahoo/maven/squidb/_latestVersion)
+
 
 ## Introducing SquiDB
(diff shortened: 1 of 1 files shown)


sbosley:
Oh nice this is cool -- do you think we should make it display the latest stable version rather than the latest beta version though? When I preview the file it shows me a download link for 3.0 beta 1.

jaredsburrows:
@sbosley Not sure. I just went here: https://bintray.com/yahoo/maven/squidb/view and clicked on the badge and it gave me the markdown.

jaredsburrows:
@sbosley So what did you decide? I see you have a PR for 3.0 into master. We should be able to add this in.

sbosley:
@jaredsburrows there's one thing that's still kind of weird-- after I published the 2.1.0 version, the latestVersion link went to that instead of the 3.0 beta version. It seems like it's the latest version by publication date, rather than by version number. Probably not a big deal, but it could be weird if we ever need to backport any fixes to 2.x Do you want to dig in a bit to see if it's possible to make it show the badge for the latest by version number rather than by publication date? If it turns out that's not possible it's no big deal, and we can just live with the possible occasional weirdness in that edge case.

jaredsburrows:
@sbosley Yeah its an incremental release based on the latest uploaded artifact dates. I am guessing 3.0 will become the new release right? 2.1.0 - 3.0. So once 3.0 is out, it will say 3.0 on the badge.

sbosley:
@jaredsburrows right, but it would revert back to showing e.g. 2.1.1 if we ever backported fixes to the 2.x branch. I guess though even in that case, we'd probably be likely to release a new 3.x build shortly thereafter, so my earlier objection is moot. Let's do it!

jaredsburrows:
@sbosley Thanks for explaining. Sounds good!

Issue #188, opened 2016-07-01
Disable generation of Id property


aminelaadhari:
I am trying to migrate to squidb but my old database uses String ids. And as squidb supports only INTEGER as primary key, the generated class has now two methods returning 'ids': the rowid and the server id. This situation is error prone because all my code is base on the server id.

Any idea how to disable the generation of the Id property ?

sbosley:
This depends a little bit on exactly what you're asking. 

If you've already declared your String server id column explicitly with a `PRIMARY KEY` constraint, this is probably not currently possible. SQLite only supports a single primary key per table, and squidb assumes it to be the rowid as a means of bookkeeping about whether or not a model object exists in the table yet. We can take a look at enhancing things in such a way that you could specify the rowid to NOT be a primary key, but I can't say for sure what changes that might require -- it'll take a bit of experimenting, as we'd still need to use the rowid for bookkeeping.

It sounds though like you might be asking more about naming. The base TableModel class that supports all table models does declare `getId()` and `setId()` methods for the long rowid, so those can't be removed or renamed -- if you're already using those method names for accessing the server id there's not much that can be done without updating the code that calls those methods.

If on the other hand those accessor methods DON'T exist yet, it will be possible to make sure that there's no naming confusion. The getter/setter/Property constant names are derived from the name of the field in your model spec, but the column name underlying them can be declared as something else using the `@ColumnSpec` annotation. So, you could declare something like this:

```java
class MyModelSpec {

    // This will make the squidb integer primary key column be named myRowId instead of _id,
    // and the generated Property for the column will be the same. However, the getId and setId
    // methods will still exist as they're in the base class rather than generated code
    @PrimaryKey
    long myRowId;

    // This will make the accessors be getServerId/setServerId, but the underlying column will
    // be named "_id" -- useful if it already exists that way.
    @ColumnSpec(name = "_id")
    String serverId;
}
```

I've been  meaning to look into supporting things other than integer primary key for a while, but it's never been a high priority. Again, whether or not what you're asking will be possible right now depends a bit on your specific use case, but I'll leave the issue open to motivate us to improve our support for different kinds of primary keys.

aminelaadhari:
Thanks a lot, it's very helpful !

> if you're already using those method names for accessing the server id there's not much that can be done without updating the code that calls those methods

This is exactly my case, so I renamed the server id before starting the migration. 

[...]

PR #192, opened 2016-07-06
Primary key enhancements

sbosley:
This PR enhances our support for primary keys. It adds a lot of new features and configurability options to table models with regards to how we handle rowid bookkeeping and primary keys. For backwards compatibility reasons, a few known limitations remain, but these can be removed in the next major version of SquiDB.

These changes were inspired by the discussion in issue #188.
***
#### Background
Up until now, SquiDB has always generated an `ID` property by default for every table model. This property is declared as an `INTEGER PRIMARY KEY`, which in other words makes it an alias for the SQLite rowid column that all tables have (see [this SQLite doc](https://www.sqlite.org/lang_createtable.html#rowid)). This PR is a step towards decoupling the idea of rowid (which we need for SquiDB bookkeeping) from the idea of primary key.

#### Changes
* The `@PrimaryKey` annotation can now annotate any field, not just `long` fields. 
  * If the annotated field is an integer type, that makes the generated property an `INTEGER PRIMARY KEY`. This field will replace the default `ID` field and be used for bookkeeping. This is basically how `@PrimaryKey works now.
  * If the annotated field is some other type, that means the user has not declared an `INTEGER PRIMARY KEY`. In this case, SquiDB will generate a `ROWID` property explicitly referencing the `rowid` column, and this column will be used for bookkeeping purposes. It will be part of the `PROPERTIES` array for the model, but not explicitly created in the table (SQLite creates the column implicitly for all tables).
* The `@TableModelSpec` annotation has a new option, `boolean noRowIdAlias()`. If this flag is set, SquiDB will not generate the `ID` property as an `INTEGER PRIMARY KEY` alias to the rowid, but instead generate a `ROWID` property explicitly referencing the rowid column. This property will be used for bookkeeping instead of `ID`. Users who wish not to declare an `INTEGER PRIMARY KEY` should use this option.
  * Note: this should probably be the default behavior in the next major version of SquiDB, at which time this flag will become unnecessary and be removed
* Users who desire a multi-column primary key can declare the key using the `tableConstraints()` field in `@TableModelSpec`. Note that this also requires enabling the `noRowIdAlias()` flag, as SQLite only allows a single PRIMARY KEY declaration when creating the table.
* If none of the above apply, SquiDB will continue to generate the `ID` property as an `INTEGER PRIMARY KEY` alias to rowid. However, it will also log a warning since users will need to explicitly declare these rowid aliases in a future major version of SquiDB. If they update their model specs to explicitly declare these aliases now, this warning will go away and things will work as they always have. Updating existing model specs to remove the warning would look like this:
```java
@TableModelSpec(...)
class MyModel {
    // This replaces the legacy ID property that was generated by default 
    // with one that is semantically identical
    @PrimaryKey
    @ColumnSpec(name = "_id")
    long id;
}
```
* The existing `getId()` and `setId()` methods in `TableModel` have been deprecated and renamed to `getRowId()` and `setRowId()` to more accurately reflect their intended purpose.

#### Caveats
* In order to maintain backwards compatibility, user-defined properties cannot be named `ID`, as their getter/setter would conflict with the deprecated TableModel `getId()` and `setId()`. 
  * This restriction does not apply if the field is an `INTEGER PRIMARY KEY`, as it then replaces the default `ID` property. 
  * This restriction will be removed in a future version of SquiDB.
* User-defined columns named `rowid` are not permitted.
* SquiDB does not support `WITHOUT ROWID` tables. (This has always been the case, but is worth mentioning explicitly).
* Users who wish to migrate _existing_ tables in order to define some other primary key or remove the default rowid alias will have to run some kind of migration to drop the `_id` column. Dropping columns in SQLite is a pain, but I don't think there's much else to be done about this.


squidb-processor/src/com/yahoo/squidb/processor/plugins/defaults/properties/TableModelSpecFieldPlugin.java
```diff
@@ -83,21 +88,43 @@ protected boolean hasPropertyGeneratorForField(VariableElement field, DeclaredTy
...
+        } else {
+            generatorClass = generatorMap.get(fieldType);
+        }
         try {
-            if (field.getAnnotation(PrimaryKey.class) != null &&
-                    BasicLongPropertyGenerator.class.equals(generatorClass)) {
-                return new BasicIdPropertyGenerator(modelSpec, field, utils);
+            BasicPropertyGenerator propertyGenerator = generatorClass.getConstructor(ModelSpec.class,
+                    VariableElement.class, AptUtils.class).newInstance(modelSpec, field, utils);
+            if ("rowid".equalsIgnoreCase(propertyGenerator.getColumnName())) {
```
jdkoren: Makes me wonder if we should check other sqlite keywords too

sbosley: Not a bad idea at all, but let's file a separate issue for it. Technically SQLite columns can be named 'rowid', in which case they mask the internal rowid, but since we need it reserved for bookkeeping, this is a SquiDB-specific restriction.

squidb-processor/src/com/yahoo/squidb/processor/plugins/defaults/properties/generators/BasicPropertyGenerator.java
```diff
@@ -110,25 +115,40 @@ protected String getColumnDefault() {
...
+        if (field != null && field.getAnnotation(PrimaryKey.class) != null) {
+            PrimaryKey primaryKeyAnnotation = field.getAnnotation(PrimaryKey.class);
+            if (!toReturn.toString().toUpperCase().contains("PRIMARY KEY")) {
+                toReturn.append(" PRIMARY KEY ");
+                if (TypeConstants.isIntegerType(getTypeForAccessors()) && primaryKeyAnnotation.autoincrement()) {
+                    toReturn.append("AUTOINCREMENT");
                 }
+            } else {
+                utils.getMessager().printMessage(Kind.WARNING, "Duplicate primary key definition in column constraints."
+                        + " Use the PrimaryKey annotation instead of declaring the constraint in ColumnSpec.");
```
jdkoren: Probably could add the `@` symbol in front of the annotation names

jdkoren:
LGTM

PR #195, opened 2016-07-13
SquidCursor type arg

sbosley:
Enable SquidCursor to expose its 'hint' model class (which should match the type arg) if it was created with one. Figured this might be a useful API for some users and all the information needed for it was readily available.


jdkoren:
LGTM

Issue #183, opened 2016-06-23
Code generator should probably warn if you pass it unrecognized options


jdkoren:
Since the code generator accepts options that modify the behavior of plugins, it would be useful if options that are unrecognized print a warning during builds. This would make it apparent if an option is extraneous, has a typo in it, or similar.

Note that code generator plugins could also accept options that are unknown to squidb; perhaps the plugins themselves can report the options they accept.

sbosley:
Closed by #203, nice work @jdkoren!

PR #198, opened 2016-07-21
Log unused squidbOptions passed to APT

jdkoren:
Addresses #183


squidb-processor/src/com/yahoo/squidb/processor/plugins/defaults/AndroidModelPlugin.java
```diff
@@ -61,6 +61,12 @@ public AndroidModelPlugin(ModelSpec<?> modelSpec, PluginEnvironment pluginEnv) {
     }
 
     @Override
+    public List<String> getSupportedOptions() {
```
sbosley: All these default plugins don't need to implement this method because all the built-in default options are explicitly added to the set above.

jdkoren: Fair point

PR #202, opened 2016-07-29
More integer primary key enhancements

sbosley:
The core goal of this change is mainly to generate named aliases of the getters and setters for INTEGER PRIMARY KEY properties. (Before the only way to access these properties was with getRowId() and setRowId(), which could be confusing). These aliases use the property name, but delegate to getRowId() and setRowId() under the hood for consistent behavior.

Much of this refactor is just code reorganization. It removes the burden of tracking the ROWID property or INTEGER PRIMARY KEY rowid alias from TableModelFileWriter and moves this responsibility into the TableModelFieldSpec plugin, which is a more natural place for it and makes it easier to achieve the getter/setter behavior. This allows us to remove a lot of special case handling and duplicated code from the file writer classes.


jdkoren:
Should the files in `squidb-gen` be gitignored?

squidb-processor/src/com/yahoo/squidb/processor/plugins/defaults/properties/TableModelSpecFieldPlugin.java
```diff
@@ -125,22 +159,139 @@ private boolean isIntegerPrimaryKey(VariableElement field, DeclaredTypeName fiel
...
+                writer.writeAnnotation(CoreTypes.OVERRIDE)
+                        .beginMethodDefinition(params)
+                        .writeStringStatement("super.setRowId(rowid)")
+                        .writeStringStatement("return this")
+                        .finishMethodDefinition();
+            }
         }
     }
 
+    // Users can subclass TableModelSpecFieldPlugin and override these methods to handle the core types differently
```
jdkoren: We should probably write some javadoc for these methods in case someone wants to do this

sbosley:
@jdkoren we can clean up the sample project in a separate PR if we want, but let's leave it as is for this one at least.

jdkoren:
LGTM

PR #203, opened 2016-08-01
Plugin options handling

jdkoren:
With this change, only options declared by SquiDB should be passed to APT in the list keyed by `squidbOptions`. Any others won't cause an error but a warning will be printed during builds.

Custom plugins should use their own option key(s) if they need to inspect APT options. This has always been the intended approach, but it was never clear; this change adds appropriate messaging to this effect.

Custom plugins should specify what option key(s) they support by using the `@SupportedOptions` annotation. This is not specifically required, but it does allow SquiDB to report those back to the toolchain so that the toolchain does not warn about seemingly unused options. Example:

```java
package com.fancy.plugins;

@SupportedOptions("fancyOptions")
public class FancyPlugin extends Plugin {
    /* ... */
}
```
```gradle
apt {
    arguments {
        squidbPlugins 'com.fancy.plugins.FancyPlugin:high'
        fancyOptions 'foo'
    }
}
```

Addresses #183. Replaces #198.

### DIFF
--- a/samples/squidb-sample-core/build.gradle
+++ b/samples/squidb-sample-core/build.gradle
@@ -30,4 +30,24 @@ dependencies {
     compile project(':squidb')
     compile project(':squidb-annotations')
     apt project(':squidb-processor')
+
+    if (!generateIOSModels()) {
+        compile project(':squidb-android')
+    }
+}
+
+def generateIOSModels() {
+    // this environment variable is set by the Settings.xcconfig for IOS builds
+    !"$System.env.SQUIDB_IOS_SRC".isEmpty() && System.env.SQUIDB_IOS_SRC != null;
+}
+
+String squidbOptionsString = ''
+if (!generateIOSModels()) {
+    squidbOptionsString += 'androidModels'
+}
+
+apt {
+    arguments {
+        squidbOptions squidbOptionsString
+    }
 }
--- a/squidb-processor/src/com/yahoo/squidb/processor/plugins/PluginEnvironment.java
+++ b/squidb-processor/src/com/yahoo/squidb/processor/plugins/PluginEnvironment.java
@@ -274,10 +274,19 @@ public class PluginEnvironment {
         Set<String> unsupportedOptions = new HashSet<>(squidbOptions);
         unsupportedOptions.removeAll(SQUIDB_SUPPORTED_OPTIONS);
 
-        if (!AptUtils.isEmpty(unsupportedOptions)) {
-            String message = String.format(UNSUPPORTED_OPTIONS_WARNING, String.join(SEPARATOR, unsupportedOptions));
-            utils.getMessager().printMessage(Diagnostic.Kind.WARNING, message);
+        if (AptUtils.isEmpty(unsupportedOptions)) {
+            return;
         }
+
+        StringBuilder sb = new StringBuilder();
+        for (String option : unsupportedOptions) {
+            if (sb.length() > 0) {
+                sb.append(SEPARATOR);
+            }
+            sb.append(option);
+        }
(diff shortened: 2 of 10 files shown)


squidb-processor/src/com/yahoo/squidb/processor/plugins/PluginEnvironment.java
```diff
@@ -212,47 +242,111 @@ private void processPlugin(String pluginName) {
     }
 
     /**
+     * Add a {@link Plugin} class to the list of known plugins
+     *
+     * @param plugin the plugin class
+     * @param priority the priority to give the plugin
+     */
+    public void addPlugin(Class<? extends Plugin> plugin, PluginPriority priority) {
```
sbosley: Shall we make this private as discussed?

squidb-processor/src/com/yahoo/squidb/processor/plugins/PluginEnvironment.java
```diff
@@ -212,47 +242,120 @@ private void processPlugin(String pluginName) {
...
+     * that the key is present, rather than checking the value for that key.
+     * <p>
+     * Note: Custom plugins should also be annotated with {@link SupportedOptions @SupportedOptions} to declare which
+     * environment options keys they support; these will be reported to the toolchain accordingly.
+     *
+     * @return true if the annotation processing options contains the given key, false otherwise
      */
-    public AptUtils getUtils() {
-        return utils;
+    public boolean hasEnvOption(String key) {
```
sbosley: I think it might be worth adding a `getEnvOptionValue(String key)` method just as a convenience.

squidb-processor/src/com/yahoo/squidb/processor/plugins/PluginEnvironment.java
```diff
@@ -212,47 +242,120 @@ private void processPlugin(String pluginName) {
...
+
+    /**
+     * Returns whether the value of the "squidbOptions" environment option contains the given string
+     *
      * @param option the option to check
      * @return true if the option is set, false otherwise
+     * @deprecated use {@link #hasSquidbOption} instead
      */
+    @Deprecated
     public boolean hasOption(String option) {
```
sbosley: Can we update the various pieces of squidb code that call this deprecated method to use the new name?

samples/squidb-sample-core/build.gradle
```diff
@@ -30,4 +30,24 @@ dependencies {
...
     apt project(':squidb-processor')
+
+    if (!generateIOSModels()) {
+        compile project(':squidb-android')
+    }
+}
+
+def generateIOSModels() {
+    // this environment variable is set by the Settings.xcconfig for IOS builds
+    !"$System.env.SQUIDB_IOS_SRC".isEmpty() && System.env.SQUIDB_IOS_SRC != null;
```
sbosley: Did you update the iOS sample to actually define this environment variable?

squidb-processor/src/com/yahoo/squidb/processor/plugins/PluginEnvironment.java
jdkoren: done

squidb-processor/src/com/yahoo/squidb/processor/plugins/PluginEnvironment.java
jdkoren: done

samples/squidb-sample-core/build.gradle
jdkoren: I pulled it straight out of Settings.xcconfig

sbosley: Ah I see, ok cool.

sbosley:
LGTM

PR #209, opened 2016-08-24
Fix deprecated default ID property declaration

### LINKED ISSUE #207: Bug on 3.1.1 : causes spaces in a constant, where it shouldn't
This is what the generated file had:

`    public static final LongProperty ID = new LongProperty(TABLE_MODEL_NAME, "_id", PRIMARY KEY AUTOINCREMENT);`

As you can see, "PRIMARY KEY AUTOINCREMENT" doesn't have "_" , and instead it has spaces.
This causes this error:
Error:(31, 92) error: ')' expected
Error:(31, 110) error: <identifier> expected

Here's the spec file:

```
@TableModelSpec(className = "ExperimentEntity", tableName = "experiments", tableConstraint = "UNIQUE(experimentId)")
public class ExperimentSpec {
    public String experimentId, experimentJson;
    public int experimentVersion, experimentLevel;
}
```

Reverting back to version 3.1.0 fixed the issue.

sbosley:
Fix an issue where the deprecated default ID property didn't correctly quote its column definition. Our test cases didn't catch this when we refactored it because we had overzealously updated all the test model specs to remove the deprecation warnings. This PR fixes the bug and updates one of our test model specs to make sure that the deprecated default ID case is tested in our test cases.

Closes #207


jdkoren:
LGTM, waiting for travis

PR #213, opened 2016-08-25
Misc cleanup

sbosley:
* Various updates to our travis config: 
  * Only downloads needed android components for each build
  * Test against Android N
  * Use j2objc 1.1 for iOS tests
* Bump build tools and support libs to latest versions. Also added documentation to the wiki about how to exclude transitive dependencies for things like `squidb-support-loader` and `squidb-recyclerview` so that you're not forced into using the exact version of the support lib that they do.
* Bump the SQLite version in `squidb-sqlite-bindings` to 3.14.1
* `squidb-reactive` didn't use anything Android-specific, so make it a jar instead of an aar


jdkoren:
LGTM

Issue #215, opened 2016-08-27
 Make fields getter/setter private 


MFlisar:
I'm always using the same naming and technique, like following (see the comment to get what I want):

```
@TableModelSpec(className="GroupedFolder", tableName="groupedFolder")
@Implements(interfaceClasses = {Serializable.class})
public class GroupedFolderEntrySpec
{
    // make the generated function "setInternalType" private, because from outside, only the ModelMethod below should be used
    private int internalDataSource;

    enum Type
    {
        New,
        Default,
        Custom
    }

    @ModelMethod
    public static void setType(GroupedFolder item, Type type) {
        item.setInternalType(type.ordinal());
    }
}
```

Can I somehow let squidb make a getter/setter private? It does not respect the private definition of my variable...

sbosley:
Two things to take note of here:
1) SquiDB actually supports enum data types for columns. We serialize them as a string by default, but if all you want is to have an enum column serialized to an int, you could do so by disabling the code generator's default enum handling (using the [disableEnumProperties option](https://github.com/yahoo/squidb/wiki/Taking-control-of-code-generation#compile-time-options)) and writing a code generation plugin modeled after [the default EnumFieldPlugin](https://github.com/yahoo/squidb/blob/master/squidb-processor/src/com/yahoo/squidb/processor/plugins/defaults/properties/EnumFieldPlugin.java). Then you could handle enums however you like.
2) If on the other hand you really wanted to make getters/setters private, you could do that with a code generation plugin as well. You could disable the default getters and setters with the [disableGettersAndSetters option](https://github.com/yahoo/squidb/wiki/Taking-control-of-code-generation#compile-time-options) and then write your own plugin to manually generate getters and setters using some other rule.

Personally I'd go with option 1 as it would probably be much less code, but both solutions should work.

sbosley:
Now that I'm thinking about it, it would probably be useful to have the option of whether to serialize enums using their int ordinal or string name be configurable, so that's something we can look into adding. Shouldn't be too hard.

MFlisar:
> We serialize them as a string by default

This means, enums would be saved to the database as a string field with the enums name? Did not know that I can use enums directly. I would really prefer an integer field for an enum though. Having the ability to set this up somehow would be very useful.

Secondly, what does speak against generating the getters/setters with the same visibility as the the one defined in the model spec class?

sbosley:
[...]

PR #217, opened 2016-08-30
Argument binding enhancements

sbosley:
This PR contains several related refactors and enhancements.

The motivation for this change partially stems from the discussion in issue #215. By default, non-primitive types such as enums were bound to queries or other SQL statements as strings by calling toString() on the value before binding. However, in the case of e.g. enums, it is totally reasonable that the user might prefer to bind those values using the int ordinal, or some other proprietary identifier. This concept could also extend beyond enums to other high level wrapper types.

The main goal of this PR was to provide an API that would allow users to easily customize how non-primitive arguments are resolved before binding them to a SQL statement. The refactor in this PR consists of the following changes:
  1. Introduce the ArgumentResolver interface for controlling how higher-level types are resolved to primitives before binding. A DefaultArgumentResolver implementation replaces `SqlUtils.resolveArgReferences`. DefaultArgumentResolver is used when the user does not specify any other strategy (the common case). DefaultArgumentResolver is also extensible so users can easily introduce handling of new types without losing the default behavior for things like AtomicReferences.
  2. Introduce a CompileContext object that is passed to SQL builders when compiling them into a SQL string. This CompileContext object replaces the old `compile(VersionCode)` APIs, which are now deprecated. The CompileContext can hold a VersionCode, an ArgumentResolver, and an arbitrary map of user-defined key-value pairs for tracking custom metadata if needed. CompileContext is constructed via a builder, so we can imagine introducing new parameters to it without needing to break the `compile()` APIs again. The old `compile(VersionCode)` APIs simply construct a CompileContext with default parameters.
  3. A custom CompileContext can be supplied at the SquidDatabase level by overriding  the new `buildCompileContext` hook. This hook allows users to modify the parameters of a CompileContext.Builder before it is returned to be used for compiling a Query or other SQL statement.
  4. Alternatively, a CompileContext can be set on an instance of Query or any other SQL statement. If set, this value would supersede the context provided by the SquidDatabase compiling the object, which would be ignored.
  5. Calling the supplied ArgumentResolver to resolve arguments now happens at the CompiledArgumentResolver level rather than the cursor factory level. This incidentally fixes an until-now unnoticed bug where higher-level arguments might not be correctly bound to non-query SQL statements.

So, for example, if a user wanted to handle enums by using ordinal() instead of name(), they would:
  1. Write a custom plugin for handling enum fields in model specs that would supersede the default enum handling.
  2. Extend DefaultArgumentResolver to unwrap enums using ordinal() instead of name(), and then supply this resolver in the `buildCompileContext` hook of their SquidDatabase.

Of course, this concept generalizes to any other high-level type, enum was just the most obvious example.

***
Other misc fixes:
* Blobs (byte[]) are now converted to raw SQL as blob literals.
* Default enum handling now correctly uses `name()` instead of `toString()` in case the enum overrides `toString()`.


squidb/src/com/yahoo/squidb/sql/ArgumentResolver.java
```diff
@@ -0,0 +1,24 @@
...
+ * lower-level primitives that are bound to the SQLite query at execution time. For example, an implementation of
+ * this interface might unwrap the contents of an AtomicReference, or convert an AtomicBoolean into an integer.
+ * <p>
+ * The default implementation of this interface that SquiDB uses can be found in {@link DefaultArgumentResolver}
+ */
+public interface ArgumentResolver {
+
+    /**
+     * Resolve the given argument into a primitive suitable for binding to a low-level SQLite statement. If the
+     * argument resolver does not map the given argument into a number, string, boolean, blob, etc., it will eventually
```
sbosley: 👍

squidb/src/com/yahoo/squidb/sql/DefaultArgumentResolver.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Default implementation of {@link ArgumentResolver} that unwraps AtomicReferences, AtomicBooleans, ThreadLocals, and
+ * Enum values. Users can extend DefaultArgumentResolver by overriding {@link #canResolveCustomType(Object)} and
+ * {@link #resolveCustomType(Object)} to handle resolving types that are not handled by DefaultArgumentResolver, or to
+ * handle one of the default types in a different way (for example, to resolve Enums using ordinals instead of names).
+ */
+public class DefaultArgumentResolver implements ArgumentResolver {
```
sbosley: My thought was that it's more flexible if the user IS able to entirely override the default resolution logic. I made the `resolveArgument` method in DefaultArgumentResolver final so users couldn't accidentally override the handling of other arg types when extending it, the idea being that if a user really wanted to do so they'd have to create an entirely separate class that implements the ArgumentResolver interface.  ArgumentResolver already documents that DefaultArgumentResolver is SquiDB's preferred implementation, maybe we could make that documentation more explicit somehow?

squidb/src/com/yahoo/squidb/sql/TableStatement.java
```diff
@@ -50,21 +50,49 @@
...
+    public final String sqlForValidation(CompileContext compileContext) {
+        SqlBuilder builder = buildSql(compileContext, true, true);
         return new CompiledArgumentResolver(builder).resolveToCompiledStatement().sql;
     }
 
+    /**
+     * Sets the {@link CompileContext} to be used when compiling this statement. If set, this value supersedes that
+     * provided by the {@link #compile(CompileContext)} method.
+     */
+    public void setCompileContext(CompileContext compileContext) {
```
sbosley: The ability to change the compile context on a one-off basis for a single statement was my original intention when I wrote this, but maybe it's not necessary or a good idea? We could always take it out and add it back in if there's ever demand for it.

squidb/src/com/yahoo/squidb/sql/DefaultArgumentResolver.java
jdkoren: Fair enough

squidb/src/com/yahoo/squidb/sql/TableStatement.java
jdkoren: resolved in 62719ec

jdkoren: Added #231 to record our thoughts on this

jdkoren:
LGTM

PR #218, opened 2016-08-30
BasicPropertyGenerator getter/setter enhancements

sbosley:
This PR adds some enhancements/new hooks to BasicPropertyGenerator that allow customizing the MethodDeclarationParams for the generated getter and setter. This could be used e.g. to generate getters/setters with different access modifiers or that throw checked exceptions. The writeGetterBody/writeSetterBody hooks have also been updated to have as an argument the params used to generate the method, so that all relevant information for generating the accessor is available.


jdkoren:
LGTM

PR #224, opened 2016-09-02
Apply enum hotfixes from PR #217

sbosley:
We should get these fixes for enums in ASAP and deal with the CompileContext changes in 217 separately. I confirmed that several test cases fail when I override toString in TestEnum, but pass again after the fixes were applied.


sbosley:
We might want to do this on a 3.1.3 branch instead of master, which has a few changes that are more suitable for a 3.2.0 release.

PR #233, opened 2016-10-07
Fix copyDatabase method

sbosley:
SquidUtilities.copyDatabase never quite worked correctly. The two major problems were that 1) it never copied the supporting files like the journal or WAL file, meaning if the DB hadn't been checkpointed after a transaction the main file might not have the changes yet, and 2) to be really safe, the exclusive lock should be acquired while copying the file so that no other thread could do anything to the DB while it was being copied. This PR fixes these shortcomings by moving `copyDatabase` into SquidDatabase itself, deprecating the SquidUtilities version, and ensuring that the method is covered in our test suite.

### DIFF
--- a/squidb/src/com/yahoo/squidb/utility/SquidUtilities.java
+++ b/squidb/src/com/yahoo/squidb/utility/SquidUtilities.java
@@ -193,17 +193,7 @@ public class SquidUtilities {
         byte[] buffer;
         int BUFFER_SIZE = 1024;
         buffer = new byte[BUFFER_SIZE];
-        while ((bytes = source.read(buffer)) != -1) {
-            if (bytes == 0) {
-                bytes = source.read();
-                if (bytes < 0) {
-                    break;
-                }
-                dest.write(bytes);
-                dest.flush();
-                continue;
-            }
-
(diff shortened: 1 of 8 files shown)


squidb/src/com/yahoo/squidb/utility/SquidUtilities.java
```diff
@@ -211,7 +205,6 @@ private static void copyStream(InputStream source, OutputStream dest) throws IOE
             }
 
             dest.write(buffer, 0, bytes);
-            dest.flush();
```
jdkoren: This line got me looking into file copying in java and now I'm thinking the body of this method could be simpler. Have a look at [this](http://stackoverflow.com/a/32652909) and let me know your thoughts

sbosley: Yeah it looks like the loop could be simplified a bit. Android doesn't offer the `Files` class though, and the channel-based version seems to be quite a bit slower, so just cleaning up the loop looks like the best option.

jdkoren: :+1:

jdkoren:
LGTM

PR #234, opened 2016-10-28
Prepare 3.2.0

sbosley:
Version bumps and changelog for a 3.2.0 release


build.gradle
```diff
@@ -9,9 +9,8 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.2.0'
-        classpath 'com.neenbedankt.gradle.plugins:android-apt:1.8'
```
jdkoren: Do we need a comment about using this line if you are below gradle 2.2.0?

jdkoren: Got it, for some reason I thought it was the readme file.

PR #235, opened 2016-11-01
Get the test suite running using the new testing support library

sbosley:
No more InstrumentationTestRunner deprecations!

Also cleans up/deletes a bunch of stuff in the test project that we didn't actually need.


jdkoren:
LGTM

PR #236, opened 2016-11-02
Deprecate VersionCode.LATEST

sbosley:
Defines a VersionCode constant explicitly in the squidb-sqlite-bindings module instead, since that's what this constant was really meant to be used for.


squidb/src/com/yahoo/squidb/utility/VersionCode.java
```diff
@@ -18,6 +18,13 @@
     public static final VersionCode V3_7_4 = new VersionCode(3, 7, 4, 0); // Default on API 14, default minimum
```
jdkoren: I wonder if we need `V3_7_4` and `V3_8_3` (`V3_7_11` is being used)

sbosley: Good question. We had defined them because they were sort of "significant" SQLite versions, the former being the min version of SQLite found in Android API level 14 and the latter being when common table expressions were introduced (which I was looking at implementing at one point but decided against). I bet even `V3_7_11` is really only needed for things like multi-row insert. Would it be cleaner to not have any of these constants declared here at all, and just allow each separate use case to define them as necessary?

jdkoren: I could go either way between having a collection of salient versions here or just making one-offs where needed. Since there only seems to be one, maybe the one-off is the better approach.

sbosley: I think it makes sense to have use-case specific version codes that are more descriptively named, so that it's easier to understand where they might be used. I've deprecated all the constants in `VersionCode` and defined a new one for multi-row insert, i.e. `Insert.SQLITE_VERSION_MULTI_ROW_INSERT`. Want to give it another look?

squidb-addons/squidb-sqlite-bindings/src/com/yahoo/squidb/sqlitebindings/SQLiteBindingsVersion.java
```diff
@@ -0,0 +1,17 @@
...
+ */
+package com.yahoo.squidb.sqlitebindings;
+
+import com.yahoo.squidb.utility.VersionCode;
+
+/**
+ * Container class for a {@link com.yahoo.squidb.utility.VersionCode} denoting the version of SQLite that is bundled
+ * with the default distribution of the squidb-sqlite-bindings project
+ */
+public final class SQLiteBindingsVersion {
```
sbosley: @jdkoren do you think a standalone container class makes sense for this, or could it live in SQLiteBindingsAdapter instead?

sbosley: I think I like it better in SQLiteBindingsAdapter, but if you object I'll revert it.

jdkoren:
LGTM

PR #237, opened 2016-11-02
Processor code cleanup

sbosley:
Minor cleanup to ModelSpecProcessor to fix a few harmless warnings:
* Fixed the "supported source version" warnings that users could see if compiling with java 1.8
* Updated apt-utils to 1.0.4, which adds more robust support for certain java 1.8 features and removes the noisy warnings about ErrorTypes when accumulating imports (which is expected behavior in SquiDB)

Also fixes some outdated documentation in the `@PrimaryKey` annotation.


jdkoren:
LGTM

PR #238, opened 2016-11-05
IN/NOT IN criterion enhancements

sbosley:
Minor enhancements to IN/NOT IN criterions to make them more robust against null arguments in the varargs case.


jdkoren:
LGTM

PR #239, opened 2016-11-07
Fix bug in containsNonNullValue

sbosley:
Fixes a bug in our implementation of `containsNonNullValue()` that would return true in the case when a non-null value existed in the database values even if that value had been superseded by putting null in the setValues. Thanks to user @Aranda for pointing this out!


jdkoren:
LGTM

PR #244, opened 2016-11-17
Better codegen error logging

### LINKED ISSUE #243: Handling of invalid column names (sql keywords)
It would be a nice feature, if the compiler does either of following if column names or similar are named with invalid names in the sense of that those names will result in some sql exception, like for example, a column name like "group".

*Side note/Idea*

I actually thought, that some automatic prefixig could solve that as well. But would not be backwards compatible though...

sbosley:
Introduces a new API for logging errors during code generation that helps us work around the "cannot find symbol" errors we see when using APTs built-in error logging.

This solution is kind of a hack but it works way, way better than what we're currently seeing. A problem we often see is that when logging an error using APTs Messager with Kind.ERROR, this error will fail the current round of annotation processing without making the compiler aware of any of the generated code. This leads to many, many "cannot find symbol <generated model>" errors, *even though* the class was in fact generated. Those errors mask the real error, making the root problem difficult to find and fix without sorting through a giant error log.

This PR leverages a second annotation processor to defer error logging to subsequent rounds, after model classes have already been generated. By saving some error state to the model spec (`modelSpec.logError()`), we can write an annotation to the generated model class containing info about the error we want to log. These annotations are then picked up in a subsequent round of annotation processing by ErrorLoggingProcessor -- but since they're inside the generated model class, the compiler is forced to know about the generated model class in order to process them, eliminating the "cannot find symbol" errors. The ErrorLoggingProcessor then parses the annotations to log the relevant errors. Note also that the metadata written to this error annotation contains enough info for the ErrorLoggingProcessor to actually look up the original offending element, so we can still link the error back to its actual cause.

This approach seems to work much better than using Messager in all my testing, but in case it causes problems, it's disableable using an option. I think it'll be a useful API going forward for all kinds of additional validation in the code generation step; most immediately I can push up a branch that fixes #243 after this is in.


sbosley:
Travis emulators are still misbehaving, but I ran all the tests locally and all is green, so this should be good to go.

squidb-processor/src/com/yahoo/squidb/processor/data/ErrorInfo.java
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Tuple class to hold logged error info, to be written by the
+ * {@link com.yahoo.squidb.processor.plugins.defaults.ErrorLoggingPlugin}.
+ */
+public final class ErrorInfo {
+
+    public final DeclaredTypeName errorClass; // The class on which to log the error
+    public final String element; // The specific element on which to log the error (or null/empty to log on the class)
+    public final String message; // The error message to log
```
jdkoren: Since these members are public, maybe the line comments could instead be javadoc like on `Field.NULL`? Although I suppose there's only a very small chance anyone will be using this directly over `modelSpec.logError()`.

squidb-processor/src/com/yahoo/squidb/processor/plugins/defaults/ErrorLoggingPlugin.java
```diff
@@ -0,0 +1,123 @@
...
+    @Override
+    public void addRequiredImports(Set<DeclaredTypeName> imports) {
+        if (modelSpec.getLoggedErrors().size() > 0) {
+            imports.add(MODEL_GEN_ERRORS);
+            imports.add(MODEL_GEN_ERROR_INNER);
+        }
+    }
+
+    @Override
+    public void emitAdditionalJava(JavaFileWriter writer) throws IOException {
```
jdkoren: Interesting. I expected the annotations would go on the class itself, but given how the writer works, this makes sense (and nobody should have to look at it in source anyway since they're printed by the build tools).

squidb-processor/src/com/yahoo/squidb/processor/data/ErrorInfo.java
sbosley: Sure, why not!

jdkoren:
LGTM

PR #245, opened 2016-11-21
Identifier validation

### LINKED ISSUE #243: Handling of invalid column names (sql keywords)

sbosley:
Implements validation of table, view, and column names. If the name is a keyword that is not permitted, an error will be logged during code generation. Using any other keyword will log a warning.

Closes #243


squidb-processor/src/com/yahoo/squidb/processor/SqlUtils.java
```diff
@@ -0,0 +1,203 @@
...
+    /**
+     * @return true if word is a SQLite keyword
+     */
+    public static boolean isKeyword(String word) {
+        return word == null || ALL_KEYWORDS.contains(word.toUpperCase());
+    }
+
+    /**
+     * @return true if word is a restricted SQLite keyword, i.e. cannot be used as a table or column name.
+     * If a word is a SQLite keyword, it *may* be used as a table or column name, but it is highly recommended that
```
jdkoren: Might want to copy or move this sentence to the javadoc for `isKeyword()`

sbosley:
Tests are all green on my box.

squidb-processor/src/com/yahoo/squidb/processor/plugins/defaults/properties/generators/BasicPropertyGenerator.java
```diff
@@ -67,8 +67,6 @@ public BasicPropertyGenerator(ModelSpec<?> modelSpec, VariableElement field, Apt
     private void validateColumnName() {
         if (columnName.indexOf('$') >= 0) {
```
jdkoren: I wonder if we should move this to `checkIdentifier()` so it's enforced on tables and views, too.

sbosley: Yeah that's probably a good idea.

jdkoren:
LGTM

PR #249, opened 2016-12-05
Restore older API level emulator tests

jdkoren:
(#248)
Opened so we can just re-run CI periodically until it passes.


sbosley:
We should remove the -no-audio argument in scripts/before-test.sh. It won't fix all the emulator problems (I've tried) but it is at least *one* of the issues those emulators seem to be having.

PR #250, opened 2016-12-08
Remove deprecations [4.0]

sbosley:
This PR kicks off our 4.0 development branch by removing all the deprecated methods, fields, and behavior from the 3.x branch. Specifically:
* `SquidDatabase.onDowngrade` now returns false by default
* Removes all the various generated methods and fields for the `ID` field in table models. All models in 4.0 should be using either `rowid` or an explicitly declared primary key field. Also removes restrictions on user properties named `ID` -- any user-declared property, even a non-primary key one, can now be named `ID`. Only `ROWID` is reserved.
* Removes deprecated SQL compilation methods that used `VersionCode` instead of `CompileContext`
* Other misc deprecated methods/fields that have either been removed, renamed, or refactored into better alternatives


squidb-annotations/src/com/yahoo/squidb/annotations/PrimaryKey.java
```diff
@@ -15,12 +15,9 @@
...
  * Integer-type columns (long, int, etc.) that are annotated with the PrimaryKey annotation are considered
  * "INTEGER PRIMARY KEY" columns. In SQLite, an INTEGER PRIMARY KEY column is an alias for the built-in SQLite rowid
- * column. If present, SquiDB will use such a column for model bookkeeping. If a non-integer column is annotated with
- * PrimaryKey, SquiDB will generate an explicit reference to the built-in rowid column to use for bookkeeping.
- * <p>
- * If a PrimaryKey annotation is not present in a TableModelSpec class, a default INTEGER PRIMARY KEY ID property for
- * the table will automatically be generated. However, this behavior will be removed in a future version, in favor
- * of using the built-in rowid column directly when no other primary key is declared.
+ * column. If present, SquiDB will use such a column for model bookkeeping. If no column is annotated with @PrimaryKey,
+ * or if a a non-integer column is annotated with @PrimaryKey, SquiDB will generate an explicit reference to the
```
jdkoren: nit: extra 'a'

sbosley:
Definitely yes, I was planning to write a bunch of test cases that cover the migration from 3.x to 4.0 (and I also think we should release a 4.0 beta for people to experiment with and report any issues). I'll fix that extra 'a' in a future PR.

PR #252, opened 2016-12-09
Fix clone()

### LINKED ISSUE #251: Model clone method is not working correctly
I think here's a small bug:

https://github.com/yahoo/squidb/blob/cf35046974e2cf7ef6d1f727c9aaf72d9780275f/squidb/src/com/yahoo/squidb/data/AbstractModel.java#L171

The problem is, this clone method does not create a new `transitoryData`, so the clone and the base `Model` operate on the same `HashMap` object...

sbosley:
Fixes bad handling of transitory values in `AbstractModel.clone()`

Closes #251


jdkoren:
LGTM

jdkoren:
Changes LGTM

sbosley:
Update: in auditing the transitory code I noticed they would behave badly when using a null key or more importantly null values. Added fixes for that as well.

jdkoren:
Changes still LGTM

PR #253, opened 2016-12-10
Prepare 3.2.3

sbosley:
Prepare 3.2.3 release to fix transitory bugs


jdkoren:
LGTM

sbosley:
I noticed one or two other minor bugs with transitory value handling this weekend, so I'll push up fixes and new test cases for them tomorrow and then we can get a 3.2.3 out.

PR #255, opened 2016-12-16
Refactor PropertyGenerator API [4.0]

sbosley:
Rewrites the PropertyGenerator API to make a clearer distinction between property generators for each different kind of model spec. Each has its own interface with model-specific methods that extend from a base interface common to all property generators. The field handling plugins have been rewritten to accommodate the new API.

Each different type of ModelSpec wrapper class (TableModelSpecWrapper, ViewModelSpecWrapper, etc.) now only accepts PropertyGenerator instances that implement the interface corresponding to that type of model (TableModelPropertyGenerator, ViewModelPropertyGenerator).

These changes have been further refined in minor ways by some refactoring in a future PR that I haven't opened yet, but let's review this base change on its own first.


jdkoren:
LGTM

PR #258, opened 2016-12-20
First-class support for upsert() [4.0]

sbosley:
Implements a higher-level method in SquidDatabase for "upsert", aka update or insert. Upsertable models define a "logical key" -- a tuple of non-null columns (maybe the same as the primary key) that uniquely identify a row in the database. When persisting a model using upsert, the method will first check to see if there is an existing model in the database matching the logical key values. If it exists, an update with the changed values will be performed; otherwise, the model will be inserted.

Our implementation is supported by a new 'Upsertable' interface and and `@UpsertKey` annotation for code generation to take care of building the logical key for a model into a criterion. A few pieces of behavior are configurable using the `@UpsertConfig` annotation -- whether or not to ignore the model rowid when upserting if one is present, and whether or not to throw a runtime exception or return false if the model being upserted doesn't have values for any of the logical key columns.

Currently the code generator generates a unique Index that the user is required to create in the database to enforce the uniqueness constraint. In a future enhancement I'd prefer to get rid of this index and simply do some regex matching on the table constraint to make sure that the user has declared either a `UNIQUE` or a `PRIMARY KEY` constraint for the columns -- IMO it's better to log an error during compilation than risking letting the user forget a required uniqueness constraint and potentially end up with bad data in the db. I'll make this enhancement before we put out a 4.0 beta, but let's get this basic foundation in first.


squidb-annotations/src/com/yahoo/squidb/annotations/UpsertConfig.java
```diff
@@ -0,0 +1,33 @@
...
+     */
+    boolean rowidHasPriority() default true;
+
+    /**
+     * Configures the behavior of the Upsertable.getUpsertKeyLookupCriterion() implementation for the generated model.
+     * If the model instance passed to upsert is not populated with the values necessary to construct a matching
+     * criterion, a value of true for this field will fail loudly with a runtime exception. A value of false will
+     * make upsert fail quietly by simply returning false to indicate the upsert failed. Default is true.
+     */
+    boolean lookupCriterionFailLoudly() default true;
```
jdkoren: This may be a nit, but maybe "FailThrowsException"? Or maybe "FailSilently" with default = false?

squidb-processor/src/com/yahoo/squidb/processor/plugins/defaults/UpsertPlugin.java
```diff
@@ -0,0 +1,279 @@
...
+
+        writer.writeStatement(Expressions.fromString("result").returnExpr());
+    }
+
+    private Expression badValueFailureExpression(final String columnRef, boolean failureThrowsException) {
+        if (failureThrowsException) {
+            return new Expression() {
+                @Override
+                public boolean writeExpression(JavaFileWriter writer) throws IOException {
+                    writer.appendString("throw new IllegalArgumentException(\"Value for upsert logical key column \""
```
jdkoren: I wonder if IllegalStateException is better here?

squidb/src/com/yahoo/squidb/data/Upsertable.java
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * If an Upsertable model returns true for this method, and upsert() is called with a model that has a rowid set,
+     * the logical upsert key will be ignored and upsert() will attempt only an update operation on the row with the
+     * given id. By default, the upsert code generation will generate an implementation that returns true, and
+     * upsert should generally be called with models that have no rowid set.
+     *
+     * @return true if the Upsertable model should use the rowid if one is set when calling upsert(), false if upsert
+     * should always insert/update data (including the rowid) using the logical upsert key rather than the rowid.
+     */
+    boolean rowidHasPriority();
```
jdkoren: "rowidHasUpsertPriority()" might be more clear

squidb-processor/src/com/yahoo/squidb/processor/plugins/defaults/UpsertPlugin.java
```diff
@@ -0,0 +1,279 @@
...
+                } else {
+                    return 0;
+                }
+            }
+        });
+        checkUpsertColumnOrder();
+        checkUpsertColumnsForNotNullConstraints();
+    }
+
+    // Make sure that the order values were reasonable, i.e. monotonically increasing
```
jdkoren: Not sure this is really necessary, I might be okay with just letting equally-ordered elements just be ordered indeterminately among themselves.

sbosley: This will be going away in a future enhancement before we release anyways -- just want to get the other code generation refactors in first before I rebuild this part.

squidb-processor/src/com/yahoo/squidb/processor/plugins/defaults/UpsertPlugin.java
sbosley: Agreed, since it could be called outside the context of the upsert method.

PR #259, opened 2016-12-21
Use JavaPoet for the code generator [4.0]

sbosley:
My old apt-utils library is sort of klugey and a pain to maintain. Square's JavaPoet is the new hotness, has an awesome API, and is actively under development. Let's use it instead!

This PR rewrites the entirety of the code generator to use javapoet instead of apt-utils. Most notably, this changes the plugin API in some pretty significant ways. JavaPoet prefers building up an entire specification of the generated class before writing code, so as such the "before/after" hooks don't apply in quite the same way since JavaPoet determines the ultimate order in which things in the spec are written to the generated class. Fortunately they choose a pretty reasonable ordering, so with some good use of static initializers we can get essentially the same behavior that we had with our old "append as you go" approach.

This is a big change, but JavaPoet is awesome and way more powerful than apt-utils ever was, so I think it'll let us do even more cool things going forward. In the most immediate future I'm hoping it will help support adding nullability annotations to our generated code.


jdkoren:
LGTM

PR #261, opened 2017-01-09
Plugin init

sbosley:
This PR introduces a few refactors to the Plugin and ModelSpec APIs:
* Plugins are now required to have a 0-arg constructor, and are initialized via an `init()` method. A plugin that does not want to handle a particular model spec can return false from the init method, and it will be discarded. This replaces `hasChangesForModelSpec()` makes it easier/safer for a plugin to initialize its internal state outside of its constructor, and then only if it wants to.
* `Plugin` itself is now an interface, with `AbstractPlugin` provided as a stub class to extend. This is intended to mirror how the `Processor/AbstractProcessor` API is designed.
* `ModelSpec` and its associated subclasses are now constructed via a factory instead of directly (making their constructors hidden). Some business logic has been moved from the `ModelSpec` constructor into an `initialize()` method, which is called by the factory. This will make it easier to write better tests for model specs someday.
* `will/did` naming scheme in plugin has been reverted to the old `before/after` naming scheme.


squidb-processor/src/com/yahoo/squidb/processor/plugins/defaults/enums/EnumPluginBundle.java
```diff
@@ -16,10 +15,10 @@
...
 public class EnumPluginBundle extends PluginBundle {
 
-    public EnumPluginBundle(ModelSpec<?, ?> modelSpec, PluginEnvironment pluginEnv) {
-        super(modelSpec, pluginEnv, Arrays.asList(
-                new EnumFieldPlugin(modelSpec, pluginEnv),
-                new EnumViewPropertyPlugin(modelSpec, pluginEnv),
-                new EnumInheritedPropertyPlugin(modelSpec, pluginEnv)));
+    public EnumPluginBundle() {
+        super(Arrays.asList(
+                (Class<? extends Plugin>) EnumFieldPlugin.class,
```
jdkoren: Same as the Json plugin

squidb-addons/squidb-json/squidb-json-compiler/src/com/yahoo/squidb/json/JSONPlugin.java
```diff
@@ -16,10 +15,10 @@
...
 public class JSONPlugin extends PluginBundle {
 
-    public JSONPlugin(ModelSpec<?, ?> modelSpec, PluginEnvironment pluginEnv) {
-        super(modelSpec, pluginEnv, Arrays.asList(
-                new JSONFieldPlugin(modelSpec, pluginEnv),
-                new JSONViewPropertyPlugin(modelSpec, pluginEnv),
-                new JSONInheritedPropertyPlugin(modelSpec, pluginEnv)));
+    public JSONPlugin() {
+        super(Arrays.asList(
+                (Class<? extends Plugin>) JSONFieldPlugin.class,
```
jdkoren: This cast looks a little funky here. The alternative is

```java
Arrays.<Class<? extends Plugin>>asList(
    JSONFieldPlugin.class,
    JSONViewPropertyPlugin.class,
    JSONInheritedPropertyPlugin.class
);
```

which also looks a little funky, but is probably more clear.

sbosley: Oh you're right, that's much better.

jdkoren:
LGTM

PR #265, opened 2017-01-25
Prepare 4.0 beta 1

sbosley:
Let's get an early beta of 4.0 out so people can test the new code generator, upsert, and other changes.


squidb/src/com/yahoo/squidb/data/Upsertable.java
```diff
@@ -11,7 +11,8 @@
 /**
  * Interface that can be applied to a generated {@link TableModel} class to make the model compatible with
  * {@link SquidDatabase#upsert(TableModel)}. Upsertable models are generally uniquely identified in their table by a
- * column or collection of columns referred to as a "logical key".
+ * column or collection of columns referred to as a "logical key". Logical keys can be specified in a model spec
+ * using the &#064;UpsertKey annotation.
```
jdkoren: Maybe we should use an {@link} for this?

sbosley: I think when I did that it was an error because the squidb core module doesn't have a dependency on `squidb-annotations`. I didn't try hard to find a way around that though.

jdkoren: Okay never mind then

PR #268, opened 2017-02-02
Model method enhancements [4.0]

sbosley:
Some enhancements to `ModelMethodPlugin`, including:
* Annotations on methods and method parameters are now copied to the generated model class. The `@ObjectiveCName` annotation on model instance methods is specially handled to adjust the name and remove the first argument.
* The first arg (i.e. model instance) required by methods annotated with `@ModelMethod` is now subject to more robust validation.


squidb-processor/src/com/yahoo/squidb/processor/plugins/defaults/ModelMethodPlugin.java
```diff
@@ -42,26 +51,110 @@
...
+                staticModelMethods.add(e);
+            } else if (!modifiers.contains(Modifier.PRIVATE)) {
+                pluginEnv.getMessager().printMessage(Diagnostic.Kind.WARNING,
+                        "This method will not be added to the model definition. " +
+                                "Did you mean to annotate this method with @ModelMethod?", e);
+            }
+        } else {
+            List<? extends VariableElement> params = e.getParameters();
+            if (params.size() == 0) {
+                modelSpec.logError("@ModelMethod methods must have an abstract model as their first argument", e);
```
jdkoren: I feel like this one and the one below could be the same message

Issue #242, opened 2016-11-09
Specify nullability type with package-info.java files


squall2022:
Using j2objc and Swift 3 generate a lot of warnings because of nullability not specified, like explained here : https://groups.google.com/forum/#!topic/j2objc-discuss/zNt2eMcLvlY

Addind package-info.java files to squidb would be very helpful, see Tom Ball suggestion :

_For your project's Java sources, you can define the correct nullability annotations and j2objc will translate them into Objective C annotations (be sure and include jsr305.jar to the j2objc classpath). This has the advantage of reducing the amount of casting that needs to be done when accessing Java classes in Swift. The Guava team recently did this, and simplified that effort by [putting the default annotations in each package's package-info.java](https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/package-info.java#L225) files (creating them, in many cases). That way, separate nullability annotations only needed to be added to individual source files when their API deviated from those defaults._

Thanks

sbosley:
We're working on adding nullability annotations to the SquiDB API (see #155 and #241). I'm not sure the package-info solution would do much in the short term, because neither the `@ParametersAreNullableByDefault` or the `@ParametersAreNonnullByDefault` are really "correct" for SquiDB on their own -- we'd need to add nullability annotations to the rest of the API for it to be correct, right? We're not using Swift with SquiDB yet so I'm not really familiar with what the implications here are.

sbosley:
Is the problem specifically that _Null_unspecified causes warnings, or that j2objc doesn't generate _Null_unspecified in the Swift interoperability code?

squall2022:
Yes, the correct way would be to add nullability annotations to the rest of the API, because not specifying it could cause runtime crashes, as explained [here](https://bugs.swift.org/browse/SR-104).
At time of this Swift issue, compiler didn’t generate warnings when nullability was nor specified, but now it is the case.
The warnings are nor really the problem, but it’s more to prevent runtime crashes.

For your second question, this is _Null_unspecified that causes warnings.

sbosley:
[...]

Issue #155, opened 2016-04-07
Request: add annotations of @Null and @NonNull when needed


AndroidDeveloperLB:
As an example:

final SquidCursor<SomeEntity> cursor = fetchFirstItem(SomeEntity.class, query);

The returned value is always not-null, because "fetchFirstItem" calls "query" , which always returns "new SquidCursor" .

So, you could make it this way:

    @NotNull
    protected <TYPE extends AbstractModel> SquidCursor<TYPE> fetchFirstItem(Class<TYPE> modelClass, Query query) {

This can help with handling possible errors.

sbosley:
We can consider this, but I think it's unlikely we'll be able to add these since my hunch is that they wouldn't be supported by j2objc on our 3.0 dev branch. In 3.0, the main squidb module is a jar that needs to be free of android dependencies, and these annotations are found in the android support annotations library.

AndroidDeveloperLB:
But this repo is only for Android, no?
Isn't there a way to overcome this?

At least, add Javadocs to tell for each function its rules of parameters
and returned values...

sbosley:
The repo is android-only at the moment on the master branch, but it won't be when the dev_3.0 branch is merged. We're moving towards multi-platform support instead of being an Android-only library.

Javadocs are an option. If you see places where you think the documentation could be more detailed feel free to open a pull request!

AndroidDeveloperLB:
Every function should have Javadocs, especially those that are most used.

sbosley:
This is being worked on using the JSR305 nullability annotations.

sbosley:
We've just released version 4.0 beta 2, which adds jsr305 `@Nullable` and `@Nonnull` to the squidb public API and generated code. If you'd like to try out the latest beta, see the README on [this branch](https://github.com/yahoo/squidb/tree/dev_4.0).

AndroidDeveloperLB:
@sbosley Are there any changes needed in gradle/code for moving to this version from the one I used here:
https://github.com/yahoo/squidb/issues/263
?

sbosley:
@AndroidDeveloperLB yes to both. The 4.0 beta versions contain several changes from the 3.x versions. I'll outline a few here, but you should read the full changelog to determine if any of the other changes will affect your code.

[...]

PR #270, opened 2017-02-09
Add nullability annotations to the public API of SquiDB [4.0]

sbosley:
Refs #155 and #242 

We use the jsr305 versions of the `@Nullable` and `@Nonnull` annotations, because those are the annotations that are included in j2objc. 

Nullability annotations have been added to most of the public API, although j2objc users should take note that since the JRE emul project does not have nullability completeness, it is still best to compile SquiDB and other shared Java code with j2objcc or to use the -Wno-nullability-completeness flag for those files.

This commit does not include nullability annotations for the generated getter and setter methods in model classes. That change will be coming in a future enhancement.


squidb-processor/src/com/yahoo/squidb/processor/plugins/defaults/properties/generators/RowidPropertyGenerator.java
```diff
@@ -57,6 +58,9 @@ public TypeName getTypeForAccessors() {
     private void addAccessorDocumentationForRowids(MethodSpec.Builder params, boolean getter) {
         if (isUnaliasedRowid()) {
             params.addAnnotation(Override.class);
+            if (!getter) {
```
sbosley: No, because the getter in the unalised rowid case returns a primitive, so it doesn't need an annotation. The setter is the one that returns the table model object.

jdkoren: Ah, got it.

PR #271, opened 2017-02-11
Create Table/View SQL [4.0]

sbosley:
Add getCreateTableSql and getCreateViewSql methods to Table and View. This makes it easier for users to use our code generation/model objects for defining schemas and building SQL statements more easily without committing to using SquidDatabase as the DAO.

### DIFF
--- a/squidb/src/com/yahoo/squidb/sql/Table.java
+++ b/squidb/src/com/yahoo/squidb/sql/Table.java
@@ -7,7 +7,6 @@ package com.yahoo.squidb.sql;
 
 import com.yahoo.squidb.data.TableModel;
 import com.yahoo.squidb.sql.Property.LongProperty;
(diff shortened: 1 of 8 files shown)


squidb/src/com/yahoo/squidb/sql/Table.java
```diff
@@ -130,6 +130,18 @@ public void appendCreateTableSql(@Nonnull CompileContext compileContext, @Nonnul
     }
 
     /**
+     * @return the <code>CREATE TABLE</code> statement for creating this table. Users should generally not need to call
+     * this method directly unless they are not working with a SquidDatabase instance and wish to create tables
+     * manually.
+     */
+    @Nonnull
+    public String getCreateTableSql() {
```
sbosley: I wonder if these should take a `CompileContext` argument instead of using a fixed SQLite version code.

PR #272, opened 2017-02-21
Typesafe column default annotations [4.0]

sbosley:
The main focus of this PR is to introduce typesafe default value annotations for table model specs that replace the now-removed `@ColumnSpec(defaultValue = "?")`. There are annotations for each primitive type (`@DefaultLong`, `@DefaultString`, etc.), as well as an explicit `@DefaultNull` and `@DefaultExpression` for specifying defaults as SQL expressions (in particular adding support for the special cases like `CURRENT_TIMESTAMP`). In addition to the annotations themselves only accepting a value of the correct type, we also perform validation confirming that only the expected type is used when defining a column default (i.e. we can log an error if a long column is annotated with `@DefaultString`).

This PR also removes the `@ColumnSpec(name = "colName")` field in favor of a dedicated `@ColumnName("colName")` annotation. This is the first step in a larger effort to break apart the different kinds of column definition components into their own annotations, rather than having a single monolithic annotation like `ColumnSpec` that does too many different things. Currently, `ColumnSpec` still exists only for the `constraints` field, but I expect that to also be replaced by a dedicated annotation in a future change.


samples/squidb-ios-sample/squidb-ios-sample/squidb-gen/com/yahoo/squidb/sample/models/Tag.java
```diff
@@ -3,6 +3,7 @@
```
sbosley: No, this is in the sample project -- we check in the iOS generated files. I can't remember if there was a reason for that or if it happened accidentally; I can look into it to find out, but separate issue.

squidb-annotations/src/com/yahoo/squidb/annotations/defaults/DefaultNull.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation for explicitly specifying that a column has a default value of null. This also guarantees that null
+ * will appear in the model default values.
+ */
+@Target(ElementType.FIELD)
+public @interface DefaultNull {
```
sbosley: I know, hence the javadocs in this annotation -- this annotation is used to explicitly specify that a default value of null appears in the in-memory model default values. This is analogous to the current behavior of `@ColumnSpec(defaultValue = ColumnSpec.DEFAULT_NULL)`. SQLite uses a null default either way, but right now we need the annotation for the in-memory values. I have it on my roadmap to change this in a future PR as well, but it's out of scope for this PR.

samples/squidb-ios-sample/squidb-ios-sample/squidb-gen/com/yahoo/squidb/sample/models/Tag.java
sbosley: Looks like it did happen accidentally, I can remove them.

squidb-annotations/src/com/yahoo/squidb/annotations/defaults/DefaultBool.java
```diff
@@ -0,0 +1,18 @@
...
+package com.yahoo.squidb.annotations.defaults;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation for specifying a default value on a boolean column
+ */
+@Target(ElementType.FIELD)
+public @interface DefaultBool {
```
sbosley: 👍

squidb-annotations/src/com/yahoo/squidb/annotations/defaults/DefaultNull.java
jdkoren: Ah, right, I forgot about the defaultValues.

PR #276, opened 2017-03-26
Added toString to ValuesStorage to ease debugging in Mockito tests.

rkaercher:
When testing with mockito and expecting an entity generated from a tablespec one currently gets the following error from mockito (and most likely other expecting frameworks, too) if an argument does not match the expected values:

Actual invocation has different arguments:
myDao.save(
    MyFromSpecEntity
set values:
com.yahoo.squidb.android.ContentValuesStorage@43ccf00c
values:
null

);

This is not a real problem when running in the IDE but causes extra effort when getting such an error in a CI test. I therefore simply added a toString that outputs all fields from the ValueStorage to ease debugging.

### DIFF
--- a/squidb/src/com/yahoo/squidb/data/ValuesStorage.java
+++ b/squidb/src/com/yahoo/squidb/data/ValuesStorage.java
@@ -179,6 +179,18 @@ public abstract class ValuesStorage {
     }
 
     @Override
+    public String toString() {
+        StringBuilder result = new StringBuilder();
+        for (String key : keySet()) {
+            result.append(key);
+            result.append(": \"");
+            result.append(get(key));
+            result.append("\"\n");
+        }
+        return result.toString();
+    }
+
+    @Override
     public abstract boolean equals(@Nullable Object o);
 
     @Override
--- a/squidb/src/com/yahoo/squidb/data/ValuesStorage.java
+++ b/squidb/src/com/yahoo/squidb/data/ValuesStorage.java
@@ -180,13 +180,15 @@ public abstract class ValuesStorage {
 
     @Override
     public String toString() {
-        StringBuilder result = new StringBuilder();
-        for (String key : keySet()) {
-            result.append(key);
+        StringBuilder result = new StringBuilder(getClass().getSimpleName());
+        result.append(": {\n");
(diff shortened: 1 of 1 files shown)


squidb/src/com/yahoo/squidb/data/ValuesStorage.java
```diff
@@ -179,6 +179,18 @@ public void put(@Nonnull String key, @Nullable Object value, boolean errorOnFail
     }
 
     @Override
+    public String toString() {
+        StringBuilder result = new StringBuilder();
+        for (String key : keySet()) {
```
sbosley: Minor suggestion: you can use `valueSet` to iterate over the key/value pairs all at once without needing to call `get(key)` for each key.

squidb/src/com/yahoo/squidb/data/ValuesStorage.java
```diff
@@ -179,6 +179,18 @@ public void put(@Nonnull String key, @Nullable Object value, boolean errorOnFail
     }
 
     @Override
+    public String toString() {
+        StringBuilder result = new StringBuilder();
```
sbosley: Minor suggestion: it might be helpful to have the specific type of the ValuesStorage in there for debugging too, along with some brackets to more clearly denote where values begin and end. How about something like:
```java
StringBuilder result = new StringBuilder(getClass().getSimpleName().toString()).append(": {\n");
// for loop here
result.append("}\n");
return result.toString();
```

sbosley:
This makes sense, thanks for this! Just a couple minor suggestions but it'll be great to get this in. We're planning on releasing a second beta of 4.0 soon so we'll make sure this is included in that.

squidb/src/com/yahoo/squidb/data/ValuesStorage.java
rkaercher: Yes, great idea!

squidb/src/com/yahoo/squidb/data/ValuesStorage.java
rkaercher: Oh, one year doing only little java and I already forgot about this, shame on me!

sbosley:
LGTM too, thanks @rkaercher!

PR #287, opened 2019-02-13
build-ios-sample-code-with-j2objc2.3

a83988029:


a83988029:
The Travis CI build failed because the j2objc of CI is version 1.1
I made changes for j2objc-version-2.3

